PR #18, opened 2018-02-13
Fix CreateHiveTableAsSelectHarvesterSuite failure

weiqingy:
Change:
CreateHiveTableAsSelectHarvesterSuite failed because of some changes happened in Spark 2.3.1 branch.

Testing:
Pass the current unit tests.

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanTracker.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanTracker.scala
@@ -86,11 +86,6 @@ class SparkExecutionPlanTracker(
                   logDebug(s"LOAD DATA [LOCAL] INPATH (${c.path}) ${c.table}")
                   CommandsHarvester.LoadDataHarvester.harvest(c, qd)
 
-                // Case 6. CREATE TABLE AS SELECT
-                case c: CreateHiveTableAsSelectCommand =>
-                  logDebug(s"CREATE TABLE AS SELECT query: ${qd.qe}")
-                  CommandsHarvester.CreateHiveTableAsSelectHarvester.harvest(c, qd)
-
                 case c: CreateDataSourceTableAsSelectCommand =>
                   logDebug(s"CREATE TABLE USING xx AS SELECT query: ${qd.qe}")
                   CommandsHarvester.CreateDataSourceTableAsSelectHarvester.harvest(c, qd)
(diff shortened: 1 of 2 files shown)


weiqingy:
Hi @jerryshao ,

Could you please check this issue? 
I tested this PR in HDP cluster, there is no lineage shown in Atlas UI. So I think, besides the changes caused by spark 2.3.1, probably there are some other issues going on. Hive table entities have been created for f2 and f3, but no lineage between them.

```
scala> sql("create table f2 (a int)")
res0: org.apache.spark.sql.DataFrame = []

scala> sql("insert into table f2 values (1)")
res1: org.apache.spark.sql.DataFrame = []

scala> sql("create table f3 as  select * from f2")
res2: org.apache.spark.sql.DataFrame = []
```

weiqingy:
I tested "InsertInto" case in HDP cluster too. The following case worked. All the entities and the lineage are there.

```
scala> sql("create table f4 (a int)")
res0: org.apache.spark.sql.DataFrame = []

scala> sql("insert into table f4 values (1)")
res1: org.apache.spark.sql.DataFrame = []

scala> sql("create table f5 (a int)")
res2: org.apache.spark.sql.DataFrame = []

scala> sql("insert into table f5 select * from f4")
res3: org.apache.spark.sql.DataFrame = []
```
cc: @jerryshao

weiqingy:
Test results are here:

http://172.27.16.196:21000/index.html#!/search/searchResult?type=hive_table&searchType=basic

dongjoon-hyun:
You can revert this if you want. 

BTW, 
- It's not Spark 2.3.1. It's Spark 2.3.0 RC3.
- Why don't we have a test coverage for those cases?

jerryshao:
What is our targeting Spark version? I'm a little confused, I assume we use Spark 2.3.0, but Mingjie mentioned that Spark version should be upgraded to meet his requirement, then we changed to Spark 2.3.1.

The code to parse the plan is highly version coupled, I would suggest to stick on our version unless we have to do changes necessarily.

dongjoon-hyun:
It's 2.3.0 definitely. @weiqingy is confusingly calling `2.3.1-SNAPSHOT` as a `2.3.1`.
As we know, it just a SNAPSHOT of `branch-2.3` for now.

jerryshao:
I see, thanks for clarify.

dongjoon-hyun:
For this one, Spark reverts this breaking change.
https://github.com/apache/spark/commit/d6f5e172b480c62165be168deae0deff8062f476

Sorry, @weiqingy . I'll revert this. We can recommit this patch if this is required later.

dongjoon-hyun:
Oops. My bad. Sorry guys. It was due to a late update from hortonworks repo.
Please forget about reverting.

weiqingy:
If we revert this PR, the unit tests of CreateHiveTableAsSelectHarvesterSuite will fail. That will block other PRs.

dongjoon-hyun:
Yep.

PR #44, opened 2018-05-03
[SAC-42] Unify Spark ML and Dataframe operator into a uniform Spark Process

merlintang:
This PR is used to combine Spark ML and DF operation into the same Spark Process. 
For example: 
Give a Spark application with ML and DF job, they are recorded into a same Spark Process in Atlas. 
screen shot 2018-04-23 at 6 22 48 pm

PR #47, opened 2018-05-29
[SAC-42] Unify Spark ML and Dataframe operator into a uniform Spark Process

merlintang:
This PR is used to combine Spark ML and DF operation into the same Spark Process. 
For example: 
Give a Spark application with ML and DF job, they are recorded into a same Spark Process in Atlas. 
screen shot 2018-04-23 at 6 22 48 pm


merlintang:
This is the same https://github.com/hortonworks-spark/spark-atlas-connector/pull/44. Because I revert previous one, thus, I resubmit this PR.

PR #80, opened 2018-11-06
[SAC-79] Leverage KafkaStreamWriterFactory to avoid reflection while getting destination topic

### LINKED ISSUE #79: Leverage KafkaStreamWriterFactory to avoid reflection while getting destination topic
Currently KafkaHarvester requires specific fix to spark-sql-kafka because it extracts topic information from KafkaStreamWriter which origin class doesn't have `topic` field (it is just an only one of constructor parameter and KafkaStreamWriter is not a case class), as well as it leverages Java reflection to extract topic.

KafkaStreamWriterFactory is a case class (even in Spark master branch) and have `topic` in primary constructor, so we can apply pattern matching to extract topic. One downside of this approach is additional overhead, but in this case, it is just an object creation so I guess we could ignore that, and even we can revisit and cache them if the cost of overhead matters.

HeartSaVioR:
What is proposed in the PR
----

For now, we're extracting destination topic from KafkaStreamWriter. Unfortunately neither KafkaStreamWriter is a case class nor topic is a field, so it requires custom patch on Spark side as well, and we are also leveraging reflection to extract the value.

Hopefully KafkaStreamWriterFactory is a case class instead (Spark branch-2.4 as well as master branch), so we can leverage this to get rid of custom patch as well as reflection. The cost of createWriterFactory is tiny (case class object creation) for this case, and we can find the way to cache it once we find the cost is not ignorable.

How the patch is tested
----
Another small code block to check whether pattern matching against KafkaStreamWriterFactory works.

This closes #79

### DIFF
--- a/spark-atlas-connector/src/main/scala/org/apache/spark/sql/kafka010/atlas/KafkaHarvester.scala
+++ b/spark-atlas-connector/src/main/scala/org/apache/spark/sql/kafka010/atlas/KafkaHarvester.scala
@@ -20,46 +20,48 @@ package org.apache.spark.sql.kafka010.atlas
 import scala.collection.mutable.ListBuffer
 import org.apache.atlas.model.instance.AtlasEntity
 import org.apache.spark.sql.execution.RDDScanExec
-import org.apache.spark.sql.execution.datasources.v2.WriteToDataSourceV2Exec
-import org.apache.spark.sql.kafka010.{KafkaSourceRDDPartition, KafkaStreamWriter}
+import org.apache.spark.sql.execution.datasources.v2.{DataSourceRDDPartition, DataSourceV2ScanExec, WriteToDataSourceV2Exec}
+import org.apache.spark.sql.kafka010.{KafkaContinuousInputPartition, KafkaMicroBatchInputPartition, KafkaSourceRDDPartition, KafkaStreamWriterFactory}
 import com.hortonworks.spark.atlas.AtlasClientConf
 import com.hortonworks.spark.atlas.sql.QueryDetail
 import com.hortonworks.spark.atlas.types.{AtlasEntityUtils, external, internal}
 import com.hortonworks.spark.atlas.utils.{Logging, SparkUtils}
+import org.apache.spark.sql.execution.streaming.sources.MicroBatchWriter
+
+import scala.collection.mutable
 
 object KafkaHarvester extends AtlasEntityUtils with Logging {
   override val conf: AtlasClientConf = new AtlasClientConf
 
-  def harvest(node: KafkaStreamWriter, writer: WriteToDataSourceV2Exec,
-    qd: QueryDetail) : Seq[AtlasEntity] = {
+  def extractTopic(writer: MicroBatchWriter): Option[String] = {
+    // Unfortunately neither KafkaStreamWriter is a case class nor topic is a field.
+    // Hopefully KafkaStreamWriterFactory is a case class instead, so we can extract
+    // topic information from there.
+    // The cost of createWriterFactory is tiny (case class object creation) for this case,
+    // and we can find the way to cache it once we find the cost is not ignorable.
+    writer.createWriterFactory() match {
+      case KafkaStreamWriterFactory(tp, _, _) => tp
+      case _ => None
+    }
+  }
+
+  def harvest(targetTopic: Option[String], writer: WriteToDataSourceV2Exec,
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
Two questions:
1. Would we be better to add UT to verify this? Would I be OK to only add UT for verifying the logic with extracting topic, or need to cover all on `KafkaHarvester`?
2. Do we have some guide on manual test for spark-atlas connector? I guess the code change is pretty obvious, but if we would want to manual test I'll spend time to do the test.

spark-atlas-connector/src/main/scala/org/apache/spark/sql/kafka010/atlas/KafkaHarvester.scala
```diff
@@ -70,8 +70,8 @@ object KafkaHarvester extends AtlasEntityUtils with Logging {
       e => pDescription.append(e).append(" ")
     }
 
-    if(destTopic.isDefined) {
-      pDescription.append(") Topics written into( ").append(destTopic.get).append(" )")
+    if (targetTopic.isDefined) {
```
bikassaha: Do we need to get more information like which Kakfa cluster/master this topic belongs to.

HeartSaVioR: We can extract some more information from producer parameter like we are doing for topic, but it is pretty limited to destination topic. For source topic we encounter the root issue again: we need to access KafkaSourceRDD, but it doesn't expose consumer parameters as field, as well as KafkaSourceRDD is not a case class.

IMO topic is only the information we can extract both source and sink safely. Otherwise custom patch will be needed again.

spark-atlas-connector/src/main/scala/org/apache/spark/sql/kafka010/atlas/KafkaHarvester.scala
```diff
@@ -20,46 +20,48 @@ package org.apache.spark.sql.kafka010.atlas
...
-    val inputsEntities = tChildren.flatMap{
-      case r: RDDScanExec =>
-        r.rdd.partitions.map {
-          case e: KafkaSourceRDDPartition =>
-            val topic = e.offsetRange.topic
-            read_from_topics += topic
-            external.kafkaToEntity(clusterName, topic)
-        }.toSeq.flatten
+    val topics: Set[String] = tChildren.flatMap {
+      case r: RDDScanExec => extractSourceTopicsFromDataSourceV1(r)
```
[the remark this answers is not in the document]
dongjoon-hyun: Thanks!

spark-atlas-connector/src/test/scala/org/apache/spark/sql/kafka010/KafkaTestUtils.scala
```diff
@@ -0,0 +1,487 @@
+/*
```
HeartSaVioR: This is copied from Spark: https://github.com/apache/spark/blob/master/external/kafka-0-10-sql/src/test/scala/org/apache/spark/sql/kafka010/KafkaTestUtils.scala

I expect Kafka also provides some mechanism to run embedded cluster or another lightweight approaches, but the license will not be matter so my 2 cents it seems OK.

HeartSaVioR: Ah I realized the class can be pulled with test jar... Will remove this class.

HeartSaVioR:
Addressed UT. The patch is not yet manually tested against real cluster, which may need additional days to be done.

HeartSaVioR:
@dongjoon-hyun I just checked and found I don't have write access. Could you grant me access or merge this? Thanks in advance!

dongjoon-hyun:
@HeartSaVioR . I sent the invitation. Please check it.

HeartSaVioR:
@dongjoon-hyun Thanks for the invitation. I merged this to master using granted privilege.

HeartSaVioR:
UPDATE: Verified with HDP cluster which Spark 2.4 is applied (2.4.0.3.0.3.0-196).

screen shot 2018-11-09 at 10 26 59 pm
screen shot 2018-11-09 at 10 27 06 pm
screen shot 2018-11-09 at 10 27 39 pm

Given that the patch is verified with actual Atlas cluster, I will apply commits in master branch to branch-2.4.

dongjoon-hyun:
Thanks!

PR #81, opened 2018-11-07
[SAC-79] Leverage KafkaStreamWriterFactory to avoid reflection while getting destination topic (Spark 2.3.x)

HeartSaVioR:
* Modify the changeset of #80 to work along with Spark 2.3.x
* Also add UTs for KafkaHarvester

Please check the difference between this patch and #80 : Micro-batch source of Kafka connector relies on DSv1, while continuous source relies on DSv2. Moreover there's some difference on DSv2 (Spark 2.4 sorted out some odd things like `InternalRowXXX`.)

While I added code for continuous mode, I don't expect continuous mode to trigger listener to get executed plan (it will not be finished), so spark-atlas can't work with continuous mode.

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessor.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessor.scala
@@ -23,7 +23,6 @@ import org.apache.spark.sql.execution.datasources.{InsertIntoHadoopFsRelationCom
 import org.apache.spark.sql.execution.datasources.v2.WriteToDataSourceV2Exec
 import org.apache.spark.sql.execution.streaming.sources.InternalRowMicroBatchWriter
 import org.apache.spark.sql.hive.execution._
-import org.apache.spark.sql.kafka010.KafkaStreamWriter
 import org.apache.spark.sql.kafka010.atlas.KafkaHarvester
 import com.hortonworks.spark.atlas.{AbstractEventProcessor, AtlasClient, AtlasClientConf}
 import com.hortonworks.spark.atlas.utils.Logging
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessor.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessor.scala
@@ -25,7 +25,6 @@ import org.apache.spark.sql.execution.streaming.sources.InternalRowMicroBatchWri
 import org.apache.spark.sql.hive.execution._
 import org.apache.spark.sql.kafka010.KafkaStreamWriter
 import org.apache.spark.sql.kafka010.atlas.KafkaHarvester
-
 import com.hortonworks.spark.atlas.{AbstractEventProcessor, AtlasClient, AtlasClientConf}
 import com.hortonworks.spark.atlas.utils.Logging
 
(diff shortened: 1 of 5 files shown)


HeartSaVioR:
Please hold on reviewing: I found inconsistency between code in UT (which mimics some part of SparkExecutionPlanProcessor) and actual SparkExecutionPlanProcessor... Will address it.

I think it would be better to add test for SparkExecutionPlanProcessor for kafka streaming query as well, so that it can prevent any missing like this case. Will add it too.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessor.scala
```diff
@@ -91,17 +90,13 @@ class SparkExecutionPlanProcessor(
...
-              streamWriter match {
-                case sw: KafkaStreamWriter => KafkaHarvester.harvest(sw, r, qd)
-                case _ => Seq.empty
-              }
-            } catch {
-              case _: NoSuchMethodException =>
-                logDebug("Can not get KafkaStreamWriter, so can not create Kafka topic " +
-                  s"entities: ${qd.qe}")
-                Seq.empty
+            // Unfortunately, we cannot determine whether writer is kafka or not, before calling
```
HeartSaVioR: @dongjoon-hyun @bikassaha 
Unfortunately we can't get writer information even with reflection in Spark 2.3.x. It requires calling createInternalRowWriterFactory() to get it, which we normally expect trivial overhead, but not sure it applies all data sources.

Otherwise we can't get rid of custom patch on Spark 2.3.x. Is it OK to go?

HeartSaVioR:
Just updated. Please take a look. Thanks in advance!

dongjoon-hyun:
Could you share the result in the cluster? This one needs that.

HeartSaVioR:
Let me spin up test cluster and validate against. Will post the result after validation.

HeartSaVioR:
@dongjoon-hyun 
Verified with HDP 3.0.3.0-190 cluster. Please take a look again. Thanks in advance!

screen shot 2018-11-08 at 3 51 15 pm
screen shot 2018-11-08 at 3 51 23 pm
screen shot 2018-11-08 at 3 51 37 pm
screen shot 2018-11-08 at 3 51 43 pm
screen shot 2018-11-08 at 3 52 02 pm
screen shot 2018-11-08 at 3 52 23 pm


Command:

```
./bin/spark-shell --jars spark-atlas-connector-assembly-0.1.0-SNAPSHOT.jar --conf spark.extraListeners=com.hortonworks.spark.atlas.SparkAtlasEventTracker --conf spark.sql.queryExecutionListeners=com.hortonworks.spark.atlas.SparkAtlasEventTracker --packages org.apache.spark:spark-sql-kafka-0-10_2.11:2.3.2
```

Streaming query to run:

```
val sourceTopics = Seq("sparksource1", "sparksource2", "sparksource3").mkString(",")
val targetTopic = "sparksink"

val df = spark.readStream.format("kafka").option("kafka.bootstrap.servers", bootstrapServers).option("subscribe", sourceTopics).option("startingOffsets", "earliest").load()

df.writeStream.format("kafka").option("kafka.bootstrap.servers", bootstrapServers).option("checkpointLocation", checkpointLocation).option("topic", targetTopic).start()
```

dongjoon-hyun:
Please proceed to cherry-pick to internal repo and verify once more there.

HeartSaVioR:
@dongjoon-hyun Pulled the change into internal repo and verified again.

dongjoon-hyun:
Thanks.

PR #82, opened 2018-11-07
[SAC-79][FOLLOWUP] Add actual UT for streaming query with Kafka Source/Sink

HeartSaVioR:
This is a follow-up change what I found from #81: actual test on pattern matching in SparkExecutionPlanProcessor.

### DIFF
--- a/spark-atlas-connector/src/test/scala/org/apache/spark/sql/kafka010/atlas/KafkaHarvesterSuite.scala
+++ b/spark-atlas-connector/src/test/scala/org/apache/spark/sql/kafka010/atlas/KafkaHarvesterSuite.scala
@@ -17,13 +17,27 @@
 
 package org.apache.spark.sql.kafka010.atlas
 
+import java.nio.file.Files
+import java.util.concurrent.atomic.AtomicLong
+
+import com.hortonworks.spark.atlas.sql.QueryDetail
+import com.hortonworks.spark.atlas.types.external.KAFKA_TOPIC_STRING
+import com.hortonworks.spark.atlas.types.metadata
+import com.hortonworks.spark.atlas.utils.SparkUtils
+import org.apache.atlas.model.instance.AtlasEntity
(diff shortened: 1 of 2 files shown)


spark-atlas-connector/src/test/scala/org/apache/spark/sql/kafka010/atlas/KafkaHarvesterSuite.scala
```diff
@@ -77,145 +63,6 @@ class KafkaHarvesterSuite extends StreamTest {
...
-    override def onSuccess(funcName: String, qe: QueryExecution, durationNs: Long): Unit = {
-      queryDetails += QueryDetail(qe, executionId.getAndIncrement(), durationNs)
-    }
-
-    override def onFailure(funcName: String, qe: QueryExecution, exception: Exception): Unit = {
-      fail(exception)
-    }
-  }
-
-  test("Streaming query with Kafka source and sink") {
```
HeartSaVioR: I just removed this test since it becomes redundant.

dongjoon-hyun: Please don't remove anything in this stage.

HeartSaVioR: The new test is actually verifying the same thing (the new test covers broader code path), so the code is merely duplicated. Would we leave this as it as with allowing code duplication, or with cleaning up code dup.?

dongjoon-hyun: No. Please don't remove anything.

HeartSaVioR: OK. I've restored the existed test back.

PR #84, opened 2018-11-08
[SAC-83] Use Spark Catalog

### LINKED ISSUE #83: Use Spark Catalog
Hive Metastore provides multiple Catalogs. SAC assumes that it has own `spark` catalog which consists of fully accessible non-transactional tables. `spark` catalog is new one which is created by the following method.

https://cwiki.apache.org/confluence/display/Hive/Hive+Schema+Tool

```bash
$ schematool \
  -dbType mysql \
  -createCatalog spark \
  -catalogDescription 'Default catalog, for Spark' \
  -catalogLocation hdfs://cluster:8020/apps/spark/warehouse
```

dongjoon-hyun:

Hive Metastore provides multiple Catalogs. SAC assumes that it has own `spark` catalog which consists of fully accessible non-transactional tables. `spark` catalog is new one which is created by the following method.

https://cwiki.apache.org/confluence/display/Hive/Hive+Schema+Tool

```bash
$ schematool \
  -dbType mysql \
  -createCatalog spark \
  -catalogDescription 'Default catalog, for Spark' \
  -catalogLocation hdfs://cluster:8020/apps/spark/warehouse
```

## How was this patch tested?

UTs are modified and this is also tested with Apache Spark 2.4.0 and Atlas 1.1.0 with embedded HBase/Solr manually.

This closes #83 .

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
@@ -111,11 +111,11 @@ object external {
     Seq(kafkaEntity)
   }
 
-  // ================== Hive entities =====================
-  val HIVE_DB_TYPE_STRING = "hive_db"
-  val HIVE_STORAGEDESC_TYPE_STRING = "hive_storagedesc"
-  val HIVE_COLUMN_TYPE_STRING = "hive_column"
-  val HIVE_TABLE_TYPE_STRING = "hive_table"
+  // ================== Spark's Hive Catalog entities =====================
+  val HIVE_DB_TYPE_STRING = "spark_db"
+  val HIVE_STORAGEDESC_TYPE_STRING = "spark_storagedesc"
+  val HIVE_COLUMN_TYPE_STRING = "spark_column"
+  val HIVE_TABLE_TYPE_STRING = "spark_table"
 
   def hiveDbUniqueAttribute(cluster: String, db: String): String = s"${db.toLowerCase}@$cluster"
 
--- a/spark-atlas-connector/src/test/scala/com/hortonworks/spark/atlas/sql/InsertIntoHarvesterSuite.scala
+++ b/spark-atlas-connector/src/test/scala/com/hortonworks/spark/atlas/sql/InsertIntoHarvesterSuite.scala
@@ -30,7 +30,7 @@ import org.apache.spark.sql.execution.UnionExec
 import org.apache.spark.sql.hive.execution.InsertIntoHiveTable
 import org.scalatest.{Matchers, FunSuite}
 
-import com.hortonworks.spark.atlas.types.external
+import com.hortonworks.spark.atlas.types.{metadata, external}
 import com.hortonworks.spark.atlas.WithHiveSupport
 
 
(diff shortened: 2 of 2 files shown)


dongjoon-hyun:
Thank you for review. This is verified in Spark 2.4 on HDP. cc @HyukjinKwon .

screen shot 2018-11-08 at 3 18 43 pm

I'll cherry-pick this to `branch-2.4` and RMP.

HeartSaVioR:
@dongjoon-hyun 

Please correct me if I'm missing here - I found some issue while working on #99.

We have different attributes between two table types - "Hive" and "Spark", majorly different name on attribute for column representation: `columns` and `spark_schema`.

I guess there might be possible to also have conflict for column and so, but to simplify the issue I'll explain only on table.

The type of 'spark_table' is defined via below:

https://github.com/hortonworks-spark/spark-atlas-connector/blob/777800ae4e3f0cfc7fc741ba1978613672a2bca5/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/metadata.scala#L90-L117

`spark_schema` is the name to represent columns, which retrieves array of `spark_column`.

`sparkTableToEntities` creates entities accordingly based on above definition.

https://github.com/hortonworks-spark/spark-atlas-connector/blob/master/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/internal.scala#L100-L134

`hiveTableToEntities` is no longer correctly creating entities.

https://github.com/hortonworks-spark/spark-atlas-connector/blob/master/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala#L211-L243

It is still correct for `hive_table` but we use the type as `spark_table` all the time, so once SAC recognizes a table as hive table, SAC sets attribute for columns as `columns` and Atlas doesn't recognize it.

Below is the result on running below query against Spark 2.4 default distribution.

```
spark.sql("create table test1 (col1 integer)")
```

```
2018-11-14 15:46:26,913 WARN  - [pool-1-thread-8 - 6fe3ac7a-3395-4f1e-b34e-cb6801f9962b:] ~ mapAttributes(): invalid attribute spark_column.comment. Ignored.. (EntityGraphMapper:329)
...
2018-11-14 15:46:27,011 WARN  - [pool-1-thread-7 - 40395723-35ea-4046-9bba-79f368b7fbf1:] ~ mapAttributes(): invalid attribute spark_table.columns. Ignored.. (EntityGraphMapper:329)
```

(Please let me know when we are forcing end users to set up something we will always go to `spark_*` type.)

If we would want to let every entities belong to `spark_*` types, we would be better to just remove code for all Hive entities (to avoid any missing) and only use spark related methods in internal type.

Candidates for removal are below:

* Hive related methods in external type
* Many util methods in AtlasEntityUtils which have branch for Spark vs Hive

Please let me know if we need to consider other cases as well (like the table is already belong to Hive).

dongjoon-hyun:
It's natural why you misunderstand them. Basically, in HDP 2.6, it's designed for for `Hive` metastore catalog entity and Spark `in-memory` catalog entity. Please take a look at the test cases. In HDP 3.0,  we created a new `Hive` metastore for Spark. In short, there are

1. `Spark` catalog using in-memory catalog
2. `Spark` catalog using Hive Metastore
3. `Hive` catalog using Hive Metastore

In addition, for the column-level information, we already decided to remove them because we cannot manage them cleanly (at least SAC didn't populate correctly and SAC is unable to remove them correctly.).

BTW, please make it sure that you have a better way to communicate with me. This kind of discussion is not intended to be here, @HeartSaVioR .

PR #92, opened 2018-11-12
[SAC-91] Spark Catalog PreEvent should be handled correctly

dongjoon-hyun:

Currently, SAC is collecting some information at PreEvents like DropDatabasePreEvent. However, the database described in DropDatabasePreEvent might not exist and the main event DropDatabaseEvent will not arrive because it will fail. This PR fixes some PreEvent logic and handles explicitly as No-OPs for the others.

```scala
scala> sql("drop database sparkdb2")
2018-11-12 14:16:29 WARN  SparkCatalogEventProcessor:48 - Caught exception during parsing event
org.apache.spark.sql.catalyst.analysis.NoSuchDatabaseException: Database 'sparkdb2' not found;
...
at com.hortonworks.spark.atlas.AbstractEventProcessor$$anon$1.run(AbstractEventProcessor.scala:39)
```

## How was this patch tested?

Manual with Apache Atlas 1.1.0 and Apache Spark 2.4.0


HeartSaVioR:
Would we be better to apply defensive programming, like applying the same to drop table?

PR #100, opened 2018-11-14
HDP 2.6.5.x release 

nagarajjayakumar:

(Please fill in changes proposed in this fix)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)


nagarajjayakumar:
@HeartSaVioR - I have pushed against atlas 0.8.2 branch. because HDP 2.6.5 supports only Atlas 0.8.2
In order to avoid the HDP 2.6.5 spark core patching, i have a updated local copy of core files [ml listeners and kafka read write] and class loader does the magic. all scala style check fails because of the core class included.

Let me know how to proceed. In my local build is successful.  I have used the following command.

mvn clean install -DskipTests -Dscalastyle.skip=true

HeartSaVioR:
@nagarajjayakumar 
Could we revert the change in the branch, and push your change **as a PR** so that we can review and fix the things **before pushing into repo**?

`-DskipTests -Dscalastyle.skip=true` <= this should not be skipped. We enable Travis CI and we never merge the patch which breaks Travis build, unless the build itself is unstable.

HeartSaVioR:
I just reverted commit as we decided to go through normal step to review the code for `atlas-0.8.2` branch. Closing this.

PR #111, opened 2018-11-16
[SAC-110][SQL] Spark's Hive catalog column name should be lower cases

### LINKED ISSUE #110: Spark's Hive catalog column name should be lower cases
To be consistent with Hive, we had better create it in lower cases.

dongjoon-hyun:

To be consistent with Hive, we had better create it in lower cases.

## How was this patch tested?

Manual. This should be tested widh 
```
val ATLAS_SPARK_COLUMN_ENABLED = ConfigEntry("atlas.spark.column.enabled", "false")
```

This closes #110


dongjoon-hyun:
The only failure is the same one. I'll ignore this failure, @HeartSaVioR .
```
- Kafka source(s) to kafka sink - micro-batch query *** FAILED *** (4 seconds, 69 milliseconds)
  8 did not equal 6 (SparkExecutionPlanProcessorForStreamingQuerySuite.scala:159)
```

PR #136, opened 2018-11-26
[SAC-135] Change default client to Kafka

### LINKED ISSUE #135: Change default client to Kafka
Kafka provides durable and persistent storage which makes SAC able to provide updates even Atlas goes offline due to some reasons. Moreover, Kafka also provides enterprise security.

SAC is providing option to change the type of client as 'rest' and 'kafka'. Currently default value is 'rest', and it's worth to change default to 'kafka'.

HeartSaVioR:

This patch changes the default client to Kafka which provides better stability and security. Rest API client is still supported via setting `atlas.client.type` to `rest`.

## How was this patch tested?

Manually tested (Spark 2.4, Atlas 1.1 with embedded HBase, Solr, Kafka)

This closes #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,9 @@ Also make sure atlas configuration file `atlas-application.properties` is in the
 
 If you're using cluster mode, please also ship this conf file to the remote Drive using `--files atlas-application.properties`.
 
+Spark Atlas Connector supports two types of Atlas clients, "kafka" and "rest". You can configure which type of client via setting `atlas.client.type` to whether `kafka` or `rest`.
+The default value is `kafka` which provides stable and secured way of publishing changes. Atlas has embedded Kafka instance so you can test it out in test environment, but it's encouraged to use external kafka cluster in production. If you don't have Kafka cluster in production, you may want to set client to `rest`.
+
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
Hi @dongjoon-hyun and @HyukjinKwon 
I guess we still don't need to set up additional thing to do manual test since Atlas provides embedded Kafka. Default version of atlas-application.properties would work.

You could also track the requests from SAC to Atlas via below command. The command could be kafka-console-consumer.sh according to the installation of Kafka. Unlike HBase and Solr, Atlas doesn't include Kafka distribution so you may need to install Kafka distribution to your dev. environment to run such command.

```
kafka-console-consumer --topic ATLAS_HOOK --bootstrap-server localhost:9027 --from-beginning
```

Atlas will consume requests from Kafka and update, so you can still just check against Atlas UI page.

HeartSaVioR:
I'll merge this for now since it works well. Please ping me if the change breaks manual test. Thanks!

HeartSaVioR:
FYI: this is also ported back to `atlas-1.1` branch as well.

PR #146, opened 2018-12-06
[SAC-144] Refers catalog before creating spark_table entities when possible

### LINKED ISSUE #144: Refers catalog before creating spark_table entities when possible

HyukjinKwon:

This PR proposes to refer catalog before creating spark_table entities when possible. Looks there could be `CatalogTable` being passed as kind of a dummy in some places in Spark.

For instance, see the test added that mimics `DataFrameWriter.saveAsTable` code path. It causes missing table information as below:

screen shot 2018-12-06 at 8 38 44 pm

## How was this patch tested?

Unit test added to describe what happens and prevent regression. Manually tested as well:

screen shot 2018-12-06 at 7 33 49 pm

Closes #144

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/utils/SparkUtils.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/utils/SparkUtils.scala
@@ -90,22 +90,17 @@ object SparkUtils extends Logging {
   }
 
   /**
-   * Get the catalog table of current active SparkSession if needed.
+   * Get the catalog table of current external catalog if exists; otherwise, it returns
+   * the input catalog table as is.
    */
-  def getCatalogTableIfNeeded(tableDefinition: CatalogTable): CatalogTable = {
-    // Let's assume that input catalog table does not have full information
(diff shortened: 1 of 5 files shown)


spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/utils/SparkUtils.scala
```diff
@@ -89,6 +89,26 @@ object SparkUtils extends Logging {
     catalog
   }
 
+  /**
+   * Get the catalog table of current active SparkSession if needed.
+   */
+  def getCatalogTableIfNeeded(tableDefinition: CatalogTable): CatalogTable = {
+    // Let's assume that input catalog table does not have full information
+    // if both owner and last access time are default values (empty string and -1)
+    if (tableDefinition.owner == "" && tableDefinition.lastAccessTime == -1) {
```
dongjoon-hyun: Do we need this? I guess `try .. catch` blindly might be enough.

HyukjinKwon: yup. addressed.

PR #169, opened 2019-01-08
[SAC-168][SQL] Apply Spark catalog model definitions to Hive catalog entities

### LINKED ISSUE #167: Support tracking create external table
SAC takes care of creating table like CTAS but doesn't handle creating external table (using external source).

This issue tracks the effort to support it.

HeartSaVioR:

This patch fixes the mismatch between Hive catalog entities (attributes) and Spark catalog model definitions.
This patch also fixes the out of sync between Spark DB entity and its definition. (`ownerType`)

## How was this patch tested?

* Existing UT, and UT modified in AtlasExternalEntityUtilsSuite.scala
* Manually tested against Spark 2.4 & Atlas 1.1

The differences are shown with CTAS query:

`spark.sql("CREATE TABLE test_table_managed_sac_168_a SELECT * FROM test_table_sac_168_a")`

> before applying the patch

test_table_managed_sac_168_before_patch_spark_table

test_table_managed_sac_168_before_patch_spark_storagedesc

> after applying the patch

test_table_managed_sac_168_after_patch_spark_table

test_table_managed_sac_168_after_patch_spark_storagedesc

Please take a look at some attributes `properties`, `locationUri`, and `serde`. I guess `spark_schema` will also show up after applying the patch when enabling column.

This closes #167

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
@@ -117,6 +117,10 @@ object external {
   }
 
   // ================== Spark's Hive Catalog entities =====================
+  // Note: given that we use Spark model types for Hive catalog entities (except HWC),
+  // Hive catalog entities should follow Spark model definitions.
+  // In Atlas, the attributes which are not in definition are ignored with WARN messages.
+
   val HIVE_DB_TYPE_STRING = "spark_db"
   val HIVE_STORAGEDESC_TYPE_STRING = "spark_storagedesc"
   val HIVE_COLUMN_TYPE_STRING = "spark_column"
@@ -128,17 +132,23 @@ object external {
                        cluster: String,
                        owner: String): Seq[AtlasEntity] = {
     val dbEntity = new AtlasEntity(HIVE_DB_TYPE_STRING)
+    val pathEntity = external.pathToEntity(dbDefinition.locationUri.toString)
 
     dbEntity.setAttribute("qualifiedName",
       hiveDbUniqueAttribute(cluster, dbDefinition.name.toLowerCase))
     dbEntity.setAttribute("name", dbDefinition.name.toLowerCase)
     dbEntity.setAttribute(AtlasConstants.CLUSTER_NAME_ATTRIBUTE, cluster)
     dbEntity.setAttribute("description", dbDefinition.description)
-    dbEntity.setAttribute("location", dbDefinition.locationUri.toString)
-    dbEntity.setAttribute("parameters", dbDefinition.properties.asJava)
+
+    // FYI: commenting out attributes which are tied to Hive DB model definition
+    // dbEntity.setAttribute("location", dbDefinition.locationUri.toString)
+    // dbEntity.setAttribute("parameters", dbDefinition.properties.asJava)
+
+    dbEntity.setAttribute("locationUri", pathEntity)
+    dbEntity.setAttribute("properties", dbDefinition.properties.asJava)
     dbEntity.setAttribute("owner", owner)
-    dbEntity.setAttribute("ownerType", "USER")
-    Seq(dbEntity)
+
+    Seq(dbEntity, pathEntity)
   }
 
   def hiveStorageDescUniqueAttribute(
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
@dongjoon-hyun @HyukjinKwon 
Please take a look at this patch. I'm seeing some issues filed to rename attributes in Spark catalog to Hive one, so if we want to make sure Spark catalog models to be compatible with Hive models, the patch can be changed to opposite direction (apply Hive catalog model definitions to Spark catalog models and entities).

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -134,10 +138,14 @@ object external {
     dbEntity.setAttribute("name", dbDefinition.name.toLowerCase)
     dbEntity.setAttribute(AtlasConstants.CLUSTER_NAME_ATTRIBUTE, cluster)
     dbEntity.setAttribute("description", dbDefinition.description)
-    dbEntity.setAttribute("location", dbDefinition.locationUri.toString)
-    dbEntity.setAttribute("parameters", dbDefinition.properties.asJava)
+
+    // FYI: commenting out attributes which are tied to Hive DB model definition
+//    dbEntity.setAttribute("location", dbDefinition.locationUri.toString)
```
dongjoon-hyun: indentation?

dongjoon-hyun:
If possible, could you attach the Atlas UI screenshot after this?

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
HeartSaVioR: Thanks for pointing out. That was the default of IntelliJ behavior while commenting. Will fix.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -128,17 +132,23 @@ object external {
                        cluster: String,
                        owner: String): Seq[AtlasEntity] = {
     val dbEntity = new AtlasEntity(HIVE_DB_TYPE_STRING)
+    val pathEntity = external.pathToEntity(dbDefinition.locationUri.toString)
```
HeartSaVioR: Here I'm assuming location URI would be one of `file://` or `hdfs://`: I'm not familiar with Hive integration and not sure it can be others. 

If the value can be other than FS or HDFS, we have to discard `locationUri` attribute for Hive catalog entities, because `locationUri` now requires FS or HDFS entity (being referenced) instead of normal String (as Hive model defines for `location`).

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -156,15 +166,26 @@ object external {
       table: String,
       isTempTable: Boolean = false): Seq[AtlasEntity] = {
     val sdEntity = new AtlasEntity(HIVE_STORAGEDESC_TYPE_STRING)
+    val pathEntity = storageFormat.locationUri.map { u => external.pathToEntity(u.toString) }
```
HeartSaVioR: Same here for storage format's location URI.

HeartSaVioR:
@dongjoon-hyun 
Addressed review comments.

Btw, I left a comment regarding a silly question on Hive side: could you please help make clear on possible values on `locationUri`? Once we're clear on the possible values, we can decide whether leave or discard the `locationUri` attribute. Thanks in advance!

spark-atlas-connector/src/test/scala/com/hortonworks/spark/atlas/types/AtlasExternalEntityUtilsSuite.scala
```diff
@@ -51,7 +50,10 @@ class AtlasExternalEntityUtilsSuite extends FunSuite with Matchers with WithHive
     val dbEntity = dbEntities.head
     dbEntity.getTypeName should be (external.HIVE_DB_TYPE_STRING)
     dbEntity.getAttribute("name") should be ("db1")
-    dbEntity.getAttribute("location") should be (dbDefinition.locationUri.toString)
+    val locationUriEntity = dbEntity.getAttribute("locationUri").asInstanceOf[AtlasEntity]
```
HeartSaVioR: This should be removed once we have to discard `locationUri`.

spark-atlas-connector/src/test/scala/com/hortonworks/spark/atlas/types/AtlasExternalEntityUtilsSuite.scala
```diff
@@ -62,7 +64,7 @@ class AtlasExternalEntityUtilsSuite extends FunSuite with Matchers with WithHive
 
     val sdEntity = sdEntities.head
     sdEntity.getTypeName should be (external.HIVE_STORAGEDESC_TYPE_STRING)
-    sdEntity.getAttribute("location") should be (null)
+    sdEntity.getAttribute("locationUri") should be (null)
```
HeartSaVioR: Same here.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/internal.scala
```diff
@@ -48,7 +47,6 @@ object internal extends Logging {
     dbEntity.setAttribute("locationUri", pathEntity)
     dbEntity.setAttribute("properties", dbDefinition.properties.asJava)
     dbEntity.setAttribute("owner", owner)
-    dbEntity.setAttribute("ownerType", "USER")
```
HeartSaVioR: I just removed the attribute because we haven't defined `ownerType` for spark db model.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
dongjoon-hyun: Thank you for update. We usually follow the way Hive SAC works. Hive SAC shows the location ref with this field, doesn't it?

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/internal.scala
dongjoon-hyun: +1 for keeping this. IIRC,  Apache Hive SAC has this.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
HeartSaVioR: There're two kinds of differences: 1) name 2) type. `location` and `locationUri` has both differences. In Spark model, `locationUri` only allows `file://` and `hdfs://`, but able to create 
and link fs/hdfs entity. In Hive model, `location` is just a String, no more.

As we usually follow the way Hive Atlas connector works, then would we want to both renaming attributes and changing type of attributes to fit to Hive model? (as I commented earlier as `the patch can be changed to opposite direction`)

HeartSaVioR:
I just craft alternative version of the patch which changes spark models to be similar with hive models if there're some differences. Will close this and propose the new patch.

Issue #192, opened 2019-01-10
Support JDBC based operations


Bongani:
It appears that the SAC is unable to track spark JDBC based operations. For example:

 `val prop = new java.util.Properties
prop.setProperty("driver", "com.mysql.jdbc.Driver")
prop.setProperty("user", "root")
prop.setProperty("password", "password")
val url = "jdbc:mysql://localhost:3306/test"
val table = "employee"
val df = spark.read.format("csv").option("header", "true").load("employee.csv")
df.write.mode("append").jdbc(url, table, prop)`

Which shows the below on Atlas. It seems to know where the input came from, but no information regarding the output.

![screenshot from 2019-01-10 11-24-13](https://user-images.githubusercontent.com/1842813/50962445-8b07bc00-14ca-11e9-8aec-9526d9828108.png)

Bongani:
I have been able to code the fix for this. Will do some testing and create a pull request once I'm done.

Bongani:
I have made a pull request for the issue. If someone could kindly please review it. Here is an image showing it works now. 
 
![screenshot from 2019-01-11 13-02-10](https://user-images.githubusercontent.com/1842813/51038730-4eb08a80-15b4-11e9-8281-c04d25445959.png)

PR #200, opened 2019-01-11
Support JDBC (RDBMS) based operations

Bongani:

The SAC plugin was unable to track JDBC based operations as it was unable to identify the output, as indicated [in the following issue](https://github.com/hortonworks-spark/spark-atlas-connector/issues/192). The changes added the logic to identify a JDBC based operation and handle it. Minor refactoring was also done. And some unit tests were added to increase code coverage. Cobertura plugin was used to identify what unit tests needed to be added.

## How was this patch tested?

Unit tests were added, as well as manual testing


Bongani:
Hi @dongjoon-hyun & @HeartSaVioR, is it possible for someone to review this pull request? I'm unable to assign anyone as a reviewer so possibly the reason the pull request has yet to be noticed.

dongjoon-hyun:
Hi, @Bongani . Thank you for making PR.
Could you remove `cobertura` stuff in this PR?
If you want, you can make another PR for that.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -144,6 +147,44 @@ object external {
...
+    s"${databaseName.toLowerCase}.${tableName.toLowerCase}"
+
+  /**
+   * Gets the database in use by the RDBMS JDBC
+   *
+   * @param url The URL used by a JDBC connector
+   * @return
+   */
+  private def getRdbmsDatabaseName(url: String): String =
+    url.substring(url.lastIndexOf("/") + 1)
```
HeartSaVioR: This is easy to be fragile. JDBC URL formats are different between DBMS, as well as even it has database as last part (one of example would be mysql) it allows adding properties which will be `?k1=v1&k2=v2&...` which will be included to database name here.

I don't mean we need to support all the DBMS JDBC URLs, but I would feel safer if we support at least MySQL (MariaDB)/PostgreSQL/Oracle/MS SQL Server. (I think we need to parse URL a bit and check which RDBMS the URL represents)

It's OK if we can't get database name safely from these list: then I think we end up with simply not supporting it (might be better to log some message as DEBUG or INFO level). Unlike others, we may need to have `Option[Seq[AtlasEntity]]` as return type of `rdbmsTableToEntity` because of this.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -144,6 +147,44 @@ object external {
     Seq(dbEntity)
   }
 
+  // ================ RDBMS based entities ======================
+  val RDBMS_TABLE = "rdbms_table"
```
HeartSaVioR: Maybe we could have RDBMS_DATABASE as well? Given that we are tracking DB for Spark/Hive table as well as NAMESPACE for HBase.

spark-atlas-connector/src/test/scala/com/hortonworks/spark/atlas/types/AtlasExternalEntityUtilsSuite.scala
```diff
@@ -124,5 +122,28 @@ class AtlasExternalEntityUtilsSuite extends FunSuite with Matchers with WithHive
     pathEntity.getAttribute(AtlasClient.REFERENCEABLE_ATTRIBUTE_NAME) should be (
       tempFile.toURI.toString)
   }
+
+  test("convert jdbc properties to rdbms entity") {
+    val tableName = "employee"
+    val rdbmsEntity = external.rdbmsTableToEntity("jdbc:mysql://localhost:3306/default", tableName)
```
HeartSaVioR: Let's have various JDBC URLs as test cases to make sure what we are supporting. Even better to add some dummy properties (if JDBC URL format of such DBMS supports it) to show we are properly extracting database name.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
Bongani: Thanks for pointing this out. I had a bit of look, [and if I'm reading this correctly](https://github.com/apache/spark/blob/670bc55f8d357a5cd894e290cc2834e952a7cfe0/sql/core/src/main/scala/org/apache/spark/sql/jdbc/JdbcDialects.scala#L223), Spark currently supports 
- DB2Dialect
- DerbyDialect
- MsSqlDialect
- MySQLDialect
- OracleDialect
- PostgresDialect
- TeradataDialect

I'll write up a function to parse this, that will support the following DBMS.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
```diff
@@ -458,6 +457,9 @@ object CommandsHarvester extends AtlasEntityUtils with Logging {
     private val RELATION_PROVIDER_CLASS_NAME =
       "org.apache.spark.sql.execution.datasources.hbase.DefaultSource"
 
+    private val JDBC_PROVIDER_CLASS_NAME =
```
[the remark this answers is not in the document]
Bongani: I'm having some difficulty in understanding what would be required for this comment. Can you link to an example where this has been done?

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
Bongani: I tried doing this using rdbms_db, however during manual testing I would get an error from Atlas about the this was unknown entity. Have you experienced something similar before?

Bongani:
@HeartSaVioR I've completed the changes I was able to work on. If you could advise on the issues I had difficulty with.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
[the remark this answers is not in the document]
HeartSaVioR: I meant you can add a new object `JDBCEntities` and put your changes there, instead of adding JDBC related code to `HBaseEntities`. (So SHCEntities, HWCEntities, KafkaEntities are all of examples.) 

After then, you can add JDBCEntities into the pattern below: 

```
      case SHCEntities(shcEntities) => Seq(shcEntities)
      case HWCEntities(hwcEntities) => Seq(hwcEntities)
      case KafkaEntities(kafkaEntities) => kafkaEntities
      case JDBCEntities(jdbcEntities) => Seq(jdbcEntities)
```

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
HeartSaVioR: Please ignore my comment. We can start with only table and add it when we need.

spark-atlas-connector/src/test/scala/com/hortonworks/spark/atlas/sql/InsertIntoRdbmsHarversterSuite.scala
```diff
@@ -0,0 +1,120 @@
...
+
+    val queryDetail = testHelperQueryListener.queryDetails.last
+    planProcessor.process(queryDetail)
+    val entities = atlasClient.createdEntities
+
+    // we're expecting one file system entities,for input file
+    val fsEntities = listAtlasEntitiesAsType(entities, external.FS_PATH_TYPE_STRING)
+    assert(fsEntities.size === 1)
+
+    val tableEntity: AtlasEntity = getOnlyOneEntity(entities, "rdbms_table")
```
HeartSaVioR: Let's use `external.RDBMS_TABLE`

spark-atlas-connector/src/test/scala/com/hortonworks/spark/atlas/sql/InsertIntoRdbmsHarversterSuite.scala
```diff
@@ -0,0 +1,120 @@
...
+    assertTableEntity(tableEntity, databaseName + "." + tableName)
+
+    // check for 'spark_process'
+    val processEntity = getOnlyOneEntity(entities, metadata.PROCESS_TYPE_STRING)
+
+    val inputs = getSeqAtlasEntityAttribute(processEntity, "inputs")
+    val outputs = getSeqAtlasEntityAttribute(processEntity, "outputs")
+    val input = getOnlyOneEntity(inputs, external.FS_PATH_TYPE_STRING)
+    val output = getOnlyOneEntity(outputs, "rdbms_table")
```
HeartSaVioR: Same here.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/utils/JdbcUtils.scala
```diff
@@ -0,0 +1,130 @@
...
+    case url if url.startsWith(TERADATA_PREFIX) => getDatabaseNameTeradataFormat(url)
+    case _ =>
+      logWarn(s"Unsupported JDBC driver for url: $url")
+      ""
+  }
+
+  /**
+   * Retrieves database name where in hose:port/dbname format
+   *
+   * @param url
```
HeartSaVioR: I'd rather remove annotations if we don't explain them.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/utils/JdbcUtils.scala
```diff
@@ -0,0 +1,130 @@
...
+    val parsedUrl = url.substring(url.lastIndexOf("/") + 1)
+    if (parsedUrl.contains("?")) {
+      return parsedUrl.substring(0, parsedUrl.indexOf("?"))
+    }
+
+    parsedUrl
+  }
+
+  /**
+   * Retrieves the database name based on Teradata format
```
HeartSaVioR: Teradata -> Derby

Bongani:
@HeartSaVioR I've done the requested changes. I also updated the acceptance test scenario, as I noticed I was only testing for the scenario for using a jdbc driver for output and not input from DBMS source.

dongjoon-hyun:
Hi, @Bongani .
Your first contribution (#204) is merged, could you do the followings for this PR and your future contribution?
1. Create an issue about this.
2. Update the title of this PR by using the issue ID like `[SAC-XXX]` (please refer your first commit).
3. Update the PR description with screenshot of S3.

Bongani:
> Hi, @Bongani .
> Your first contribution (#204) is merged, could you do the followings for this PR and your future contribution?
> 
> 1. Create an issue about this.
> 2. Update the title of this PR by using the issue ID like `[SAC-XXX]` (please refer your first commit).
> 3. Update the PR description with Atlas screenshot.

@dongjoon-hyun Thanks on the advice on what needs to be done for future PRs. I have done the changes you have suggested.

spark-atlas-connector/src/test/scala/com/hortonworks/spark/atlas/sql/InsertIntoRdbmsHarversterSuite.scala
```diff
@@ -0,0 +1,97 @@
...
+    val queryDetail = testHelperQueryListener.queryDetails.last
+    planProcessor.process(queryDetail)
+    val entities = atlasClient.createdEntities
+
+    // we're expecting two table entities:
+    // one from the source table and another from the sink table
+    val tableEntities = listAtlasEntitiesAsType(entities, external.RDBMS_TABLE)
+    assert(tableEntities.size === 2)
+
+    val inputEntity = getOneEntityOnAttribute(tableEntities, "name", sourceTableName)
```
HeartSaVioR: Could we add verification for table entity's attributes, like `qualifiedName` and `database`?

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -31,14 +30,16 @@ import org.apache.hadoop.fs.Path
...
-
 import com.hortonworks.spark.atlas.{AtlasClient, AtlasUtils}
-import com.hortonworks.spark.atlas.utils.SparkUtils
+import com.hortonworks.spark.atlas.utils.{JdbcUtils, SparkUtils}
 import com.hortonworks.spark.atlas.sql.streaming.KafkaTopicInformation
 
 
 object external {
   // External metadata types used to link with external entities
+  val QUALIFIED_NAME_KEY = "qualifiedName"
```
HeartSaVioR: These keys are not only used for external, but also used for all entities. 

Maybe we would want to apply changing them to constant for all entities, but then code diff will mix up with two different things. 

I feel we can do the refactor from another PR. Until then, could we leave them as is were?

Bongani:
Hi @HeartSaVioR , I've applied the changes which you have requested. Please have a look again.

HeartSaVioR:
@Bongani 
FYI: Please raise a PR against `atlas-1.1` branch when you intended to target the patch to be compatible with Spark 2.3. It's OK if you just wanted to make it happen with Spark 2.4.

Bongani:
@HeartSaVioR Thanks for the information. For now I am targeting Spark 2.4.

PR #226, opened 2019-05-03
[SAC-225][SQL] Add harvester to handle CreateTableCommand

### LINKED ISSUE #225: Add harvester to handle CreateTableCommand
`CreateTableCommand` seems to be used when creating Hive table, like below query:

```
spark.sql("CREATE EXTERNAL TABLE IF NOT EXISTS employee_details (name STRING, age INT, emp_id INT,designation STRING) LOCATION '/employee/details'")
```

There's no harvester handling that plan.

HeartSaVioR:

This patch proposes adding harvester to handle `CreateTableCommand` which is used to create Hive table.

## How was this patch tested?

Newly added UT.

This closes #225

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
@@ -27,7 +27,7 @@ import org.apache.spark.sql.catalyst.analysis.UnresolvedRelation
 import org.apache.spark.sql.catalyst.catalog.HiveTableRelation
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.execution._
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
As noticed in UT, there're some differences between managed and external table regarding qualifier naming rule. For external table it should be same as existing Hive qualifier naming rule.

HeartSaVioR:
https://travis-ci.org/hortonworks-spark/spark-atlas-connector/builds/527815095?utm_source=github_status&utm_medium=notification

```
- file source to kafka sink - micro-batch query *** FAILED *** (374 milliseconds)
  Set() had size 0 instead of expected size 13 (SparkExecutionPlanProcessorForStreamingQuerySuite.scala:274)
```

Looks like intermittent failure. Locally passed. Rebuilding from Travis CI...

HeartSaVioR:
It looks like OK after adding mitigation code: 1605a2a . Merging.

PR #228, opened 2019-05-07
[SAC-227][SQL] Discard the logic of supporting table column entities

### LINKED ISSUE #227: Discard the logic of supporting table column entities
It has been headache and troublesome to keep the logic to support table column but having flag to disable it. Even we have dirty workaround to remove column attribute from created entities, which incurred a bug I had to fix.

We don't have any chance to turn it on, and once we support column entities I think it should be supported as default and shouldn't be able to turn it off.

Let's discard the logic and let SAC much cleaner.

HeartSaVioR:

It has been headache and troublesome to keep the logic to support table column but having flag to disable it. Even we have dirty workaround to remove column attribute from created entities, which incurred a bug I had to fix.

We don't have any chance to turn it on, and once we support column entities I think it should be supported as default and shouldn't be able to turn it off.

This patch proposes to discard the logic to supporting table column entities at all.

## How was this patch tested?

Modified UTs, also manually tested.

This closes #227


bolkedebruin:
This is a very quick merge seemingly without a lot of discussion?

HeartSaVioR:
For couple of months I was the only one maintaining SAC actively, so no one to discuss. When we turned off column tracking by default, column tracking was completely out of all tests - UTs, ITs, and manual tests. I don't think it has been working pretty well for all cases - we found hard to make column tracking correctly in sync and that's why we turned it off by default.

Things are going better - I've got some folks working with me maintaining SAC, so the discussion will be held for further changes. See #241 and #243.

havenBoy:
The function of supporting table entities is difficult to implement, but the importance is self-evident, isn’t it?
@HeartSaVioR

Issue #237, opened 2019-05-10
Disable `atlas.client.checkModelInStart` by default


HeartSaVioR:
We have changed default client as `Kafka`, but the default value of `atlas.client.checkModelInStart` is `true` which lets SAC to create REST API and try to connect with Atlas via REST API. This sounds like a contradiction.

As we provide a way to register models outside of SAC execution, we can safely disable it by default. Moreover, there's open patch for introducing Spark models in Apache Atlas (now in reviewing) so we may no longer need to have this feature after Atlas merge the patch in and have official release.

HeartSaVioR:
Closed via #238

PR #238, opened 2019-05-10
[SAC-237][CORE] Disable `atlas.client.checkModelInStart` by default

HeartSaVioR:

This patch proposes disabling `atlas.client.checkModelInStart` by default to prevent REST API to be initialized with default setting. Please refer #237 for more details on rationalization.

## How was this patch tested?

Existing UTs and ITs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,10 +54,10 @@ If you're using cluster mode, please also ship this conf file to the remote Driv
 Spark Atlas Connector supports two types of Atlas clients, "kafka" and "rest". You can configure which type of client via setting `atlas.client.type` to whether `kafka` or `rest`.
 The default value is `kafka` which provides stable and secured way of publishing changes. Atlas has embedded Kafka instance so you can test it out in test environment, but it's encouraged to use external kafka cluster in production. If you don't have Kafka cluster in production, you may want to set client to `rest`.
 
-Pre-create Atlas models
-=======================
+Pre-create Atlas models (Recommended)
+=====================================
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -70,7 +70,10 @@ The tool will leverage REST client API to request to Atlas which requires authen
 
 Given the approach is insecure regardless of security mode of cluster, we strongly encourage you to pass `--interactive-auth` as parameter, which asks you to input username and password of Atlas interactively.
 
-After running above command, you can set `atlas.client.checkModelInStart=false` in `atlas-application.properties` to skip checking and creating models in Spark Atlas Connector's startup.
+If you would like to let SAC handles it instead, you can set you can set `atlas.client.checkModelInStart=true` in `atlas-application.properties` to let SAC check and create models in Spark Atlas Connector's each startup.
+Please note that it leverages REST API even you configure `atlas.client.type=kafka`, and it would require additional configuration for authentication. (`atlas.client.username` and `atlas.client.password`)
+
+In the meanwhile, we're also working on migrating models into Apache Atlas: once it's done, we no longer require this step and leverage the models Atlas registers for the startup.
```
arunmahadevan: Does it mean any changes to the model has to go via Atlas team ? Could we just keep the way for Spark to load the models by running the command?

HeartSaVioR: According to the Atlas team, that's the standard way for dealing with models registration, and decision has been made. (It would be also easier for us to ensure the necessary models are available in Atlas.)

We might want to keep that feature for end users who wants to use the Atlas version which doesn't ship the Spark models, but maybe it's better to not keep two different approaches together which could be out of sync.

PR #241, opened 2019-05-13
[SAC-240] [CORE] Add additional attributes to spark_table

arunmahadevan:
This patch proposes to add additional attributes to spark_table viz. schema, provider, partitionProvider, partitionColumnNames.

## How was this patch tested?

* Modified existing UT to validate generated Atlas entity attribute.
* Manually tested.

Queries:

spark.range(10).write.json("/tmp/sac/json-wip-entity-deps")

1. spark.sql("CREATE TABLE test_table_wip_entity_deps_a USING json location '/tmp/sac/json-wip-entity-deps'")

Screen Shot 2019-05-14 at 11 47 06 AM

2. spark.sql("CREATE TABLE test_table_managed_wip_entity_deps_a SELECT * FROM test_table_wip_entity_deps_a")

Screen Shot 2019-05-14 at 11 47 42 AM

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
@@ -310,6 +310,12 @@ object external {
     tblEntity.setAttribute("tableType", tableDefinition.tableType.name)
     tblEntity.setAttribute("createTime", new Date(tableDefinition.createTime))
     tblEntity.setAttribute("parameters", tableDefinition.properties.asJava)
+    tblEntity.setAttribute("schemaDesc", tableDefinition.schema.simpleString)
+    tblEntity.setAttribute("provider", tableDefinition.provider.getOrElse(""))
+    if (tableDefinition.tracksPartitionsInCatalog) {
+      tblEntity.setAttribute("partitionProvider", "Catalog")
+    }
+    tblEntity.setAttribute("partitionColumnNames", tableDefinition.partitionColumnNames.asJava)
     tableDefinition.comment.foreach(tblEntity.setAttribute("comment", _))
     tableDefinition.viewText.foreach(tblEntity.setAttribute("viewOriginalText", _))
 
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/metadata.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/metadata.scala
@@ -81,7 +81,9 @@ object metadata {
     AtlasTypeUtil.createUniqueRequiredAttrDef(
       "qualifiedName", new AtlasStringType),
     AtlasTypeUtil.createOptionalAttrDef("tableType", new AtlasStringType),
+    AtlasTypeUtil.createOptionalAttrDef("schemaDesc", new AtlasStringType),
     AtlasTypeUtil.createOptionalAttrDef("provider", new AtlasStringType),
+    AtlasTypeUtil.createOptionalAttrDef("partitionProvider", new AtlasStringType),
     AtlasTypeUtil.createOptionalAttrDef(
       "partitionColumnNames", new AtlasArrayType(new AtlasStringType)),
     AtlasTypeUtil.createOptionalAttrDef(
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/internal.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/internal.scala
@@ -99,7 +99,11 @@ object internal extends Logging {
       sparkTableUniqueAttribute(db, tableDefinition.identifier.table))
     tblEntity.setAttribute("name", tableDefinition.identifier.table)
     tblEntity.setAttribute("tableType", tableDefinition.tableType.name)
-    tableDefinition.provider.foreach(tblEntity.setAttribute("provider", _))
+    tblEntity.setAttribute("schemaDesc", tableDefinition.schema.simpleString)
+    tblEntity.setAttribute("provider", tableDefinition.provider.getOrElse(""))
+    if (tableDefinition.tracksPartitionsInCatalog) {
+      tblEntity.setAttribute("partitionProvider", "Catalog")
+    }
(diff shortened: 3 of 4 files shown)


arunmahadevan:
Now the spark_table will have additional attributes like below

image

HeartSaVioR:
@arunmahadevan 

Could you add the details in the description of testing:
1) which query(-ies) did you use
2) screenshot(s)

And once you're addressing both Spark table and Hive table, you may need to create two different tables which each table fits to each case and get screenshot for each table.

Could you update the PR description based on above? Thanks in advance!

HeartSaVioR:
And even better if we can ensure both cases are covered in UTs. You're only ensuring the case when provider is "parquet".

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -310,6 +310,12 @@ object external {
     tblEntity.setAttribute("tableType", tableDefinition.tableType.name)
     tblEntity.setAttribute("createTime", new Date(tableDefinition.createTime))
     tblEntity.setAttribute("parameters", tableDefinition.properties.asJava)
+    tblEntity.setAttribute("schemaDesc", tableDefinition.schema.simpleString)
```
HeartSaVioR: Just FYI, the code will not conflict with #243, but given most of attributes being added here are not existing in Hive table model, either this or #243 should rebase and fix this once other side of patch gets merged.

arunmahadevan: I guess we need to add the attributes only for the spark_ entities ?
For the hive_ entities assuming it gets created from HMS atlas hook should we deal with that here?

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/metadata.scala
```diff
@@ -81,7 +81,9 @@ object metadata {
     AtlasTypeUtil.createUniqueRequiredAttrDef(
       "qualifiedName", new AtlasStringType),
     AtlasTypeUtil.createOptionalAttrDef("tableType", new AtlasStringType),
+    AtlasTypeUtil.createOptionalAttrDef("schemaDesc", new AtlasStringType),
     AtlasTypeUtil.createOptionalAttrDef("provider", new AtlasStringType),
+    AtlasTypeUtil.createOptionalAttrDef("partitionProvider", new AtlasStringType),
```
arunmahadevan: @HeartSaVioR , Assume this should get added to the atlas spark model JSON files as well? Let me know if that changes should go to Atlas repo.

For the hive_table entity should we just skip these attributes for now or update the hive models and HMS hook to populate this?

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
HeartSaVioR: > I guess we need to add the attributes only for the spark_ entities ?
I think "Yes".

> For the hive_ entities assuming it gets created from HMS atlas hook should we deal with that here?
These attributes are not existing in Hive models so we can just safely ignore these attributes.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/metadata.scala
HeartSaVioR: @arunmarhadevan

> Assume this should get added to the atlas spark model JSON files as well?

Yeah right. Maybe better to handle this first, and handle #243 later. cc. @aadarshjajodia You may want to track this PR as well. We'll merge this soon.

> For the hive_table entity should we just skip these attributes for now or update the hive models and HMS hook to populate this?

 I guess it's same question above - I don't think we should expand Hive models to fit to Spark catalog table's information. Let's skip these attributes.

HeartSaVioR:
@arunmahadevan 
Looks like you're pasting same snapshots for two different tables - could you also fix this?

arunmahadevan:
@HeartSaVioR updated the screenshots and test case to check attributes with hive and parquet.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/metadata.scala
aadarshjajodia: > @HeartSaVioR , Assume this should get added to the atlas spark model JSON files as well? Let me know if that changes should go to Atlas repo.

Yes Arun is correct. If we want to add any attribute to spark we should add it in the JSON file. Also I believe we can keep the attributes same as that of hive, unless there's something specific to spark which we might want to capture?

> For the hive_table entity should we just skip these attributes for now or update the hive models and HMS hook to populate this?

HeartSaVioR: @aadarshjajodia
There're so many difference of information what Hive and Spark are providing - that's why we defined Spark models instead of just leveraging Hive models. If we just let Spark entities to leverage Hive models, bunch of information will be just discarded.

Here's the excel file which compared Spark vs Hive models (slightly outdated due to the change of #169):
https://github.com/hortonworks-spark/spark-atlas-connector/files/3141841/SAC-168-table-models-description.copy.xlsx

aadarshjajodia: As mentioned by @arunmahadevan these are not needed as they would be covered in the models json

arunmahadevan: @aadarshjajodia The attributes are different so we cannot keep it same as hive.
Can you let me know if the JSON file for Spark entities are already checked into Atlas repo ? If you are yet to create it, please include the attributes defined in this PR for the `spark_table` entity.

HeartSaVioR: @aadarshjajodia
Let us just make the change here so that we are not blocked by Atlas model review. We can do it concurrently, and eventually remove the model definitions in SAC when we are 100% sure Atlas provides these by default.

Issue #242, opened 2019-05-14
Hive provider table entity and corresponding entities should reside to Hive models


HeartSaVioR:
We decided to simply assume end users use the same catalog for Spark and Hive, so that hive provider table entities (and corresponding entities) in SAC are being linked with these entities Hive Hook may create earlier.

This is basically concept of reverting #83.

PR #243, opened 2019-05-14
[SAC-242][SQL] Hive provider table entity and corresponding entities should reside to Hive models

HeartSaVioR:

Please refer #242 to see rationalization of the issue.

This patch effectively rolls back the change what #84 was introduced. SAC will leverage Hive models for Hive provider table. It doesn't only apply to Hive table entity - it also applies to Hive DB and Hive storagedesc entities as well.

## How was this patch tested?

Modified UTs, existing ITs, manually tested with local Apache Spark 2.4 & Apache Atlas 1.1:

> Query

```
spark.range(10).write.json("/tmp/sac/json-sac-242")
spark.sql("CREATE TABLE test_table_sac_242_a USING json location '/tmp/sac/json-sac-242'")
spark.sql("CREATE TABLE test_table_managed_sac_242_a SELECT * FROM test_table_sac_242_a")
```

> Screenshot

kind | screenshot
-- | --
spark_process - relationship | spark-process-relationship-inputs<br/>spark-process-relationship-outputs
spark_db | spark-db
spark_db - relationship | spark-db-relationship-table
spark_storagedesc | spark-storagedesc
spark_storagedesc - relationship | spark-storagedesc-relationship-table
spark_table | spark-table
spark_table - relationship | spark-table-relationship-db<br/>spark-table-relationship-storagedesc
hive_db | hive-db
hive_db - relationship | hive-db-relationship
hive_storagedesc | hive-storagedesc
hive_storagedesc - relationship | FIXME: somehow it does show relationship information!
hive_table | hive-table
hive_table - relationship | hive-table-relationship
fs_path | fs-path
fs_path - relationship | fs-path-relationship-process

Note that before the patch both of two tables leverage Spark models. Please refer the PR description of https://github.com/hortonworks-spark/spark-atlas-connector/pull/232

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
@@ -26,15 +26,15 @@ import com.hortonworks.spark.atlas.sql.KafkaTopicInformation
 import scala.collection.JavaConverters._
 import org.apache.atlas.AtlasConstants
 import org.apache.atlas.model.instance.AtlasEntity
-import org.apache.commons.lang.RandomStringUtils
+// import org.apache.commons.lang.RandomStringUtils
 import org.apache.hadoop.fs.Path
-import org.apache.hadoop.hive.ql.session.SessionState
+// import org.apache.hadoop.hive.ql.session.SessionState
 import org.apache.spark.sql.catalyst.catalog.{CatalogDatabase, CatalogStorageFormat, CatalogTable}
 import com.hortonworks.spark.atlas.{AtlasEntityWithDependencies, AtlasUtils}
-import com.hortonworks.spark.atlas.utils.{JdbcUtils, SparkUtils}
+import com.hortonworks.spark.atlas.utils.{JdbcUtils, Logging, SparkUtils}
 
 
-object external {
+object external extends Logging {
   // External metadata types used to link with external entities
 
   // ================ File system entities ======================
@@ -182,16 +182,41 @@ object external {
     AtlasEntityWithDependencies(kafkaEntity)
   }
 
+  // ================ RDBMS based entities ======================
+  val RDBMS_TABLE = "rdbms_table"
+
+  /**
+   * Converts JDBC RDBMS properties into Atlas entity.
+   */
+  def rdbmsTableToEntity(url: String, tableName: String): AtlasEntityWithDependencies = {
+    val jdbcEntity = new AtlasEntity(RDBMS_TABLE)
+
+    val databaseName = JdbcUtils.getDatabaseName(url)
+    jdbcEntity.setAttribute("qualifiedName", getRdbmsQualifiedName(databaseName, tableName))
+    jdbcEntity.setAttribute("name", tableName)
+
+    AtlasEntityWithDependencies(jdbcEntity)
+  }
+
+  /**
+   * Constructs the the full qualified name of the database.
+   */
+  private def getRdbmsQualifiedName(databaseName: String, tableName: String): String =
+    s"${databaseName.toLowerCase}.${tableName.toLowerCase}"
+
   // ================== Spark's Hive Catalog entities =====================
 
-  // Note: given that we use Spark model types for Hive catalog entities (except HWC),
-  // Hive catalog entities should follow Spark model definitions.
-  // In Atlas, the attributes which are not in definition are ignored with WARN messages.
+  // Note: SAC always creates Hive table and corresponding entities, whereas SAC refers
+  // existing table entity on HWC.
 
-  val HIVE_DB_TYPE_STRING = metadata.DB_TYPE_STRING
-  val HIVE_STORAGEDESC_TYPE_STRING = metadata.STORAGEDESC_TYPE_STRING
-  val HIVE_TABLE_TYPE_STRING = metadata.TABLE_TYPE_STRING
+  val HIVE_DB_TYPE_STRING = "hive_db"
+  val HIVE_STORAGEDESC_TYPE_STRING = "hive_storagedesc"
+  val HIVE_TABLE_TYPE_STRING = "hive_table"
 
+  // scalastyle:off
+  // This is based on the logic how Hive Hook defines qualifiedName for Hive DB (borrowed from Apache Atlas v1.1)
+  // https://github.com/apache/atlas/blob/release-1.1.0-rc2/addons/hive-bridge/src/main/java/org/apache/atlas/hive/bridge/HiveMetaStoreBridge.java#L833-L841
+  // scalastyle:on
   def hiveDbUniqueAttribute(cluster: String, db: String): String = s"${db.toLowerCase}@$cluster"
 
   def hiveDbToEntity(
@@ -212,36 +237,12 @@ object external {
     AtlasEntityWithDependencies(dbEntity)
   }
 
-  // ================ RDBMS based entities ======================
-  val RDBMS_TABLE = "rdbms_table"
-
-  /**
-   * Converts JDBC RDBMS properties into Atlas entity
-   *
-   * @param url
-   * @param tableName
-   * @return
-   */
-  def rdbmsTableToEntity(url: String, tableName: String): AtlasEntityWithDependencies = {
-    val jdbcEntity = new AtlasEntity(RDBMS_TABLE)
-
-    val databaseName = JdbcUtils.getDatabaseName(url)
-    jdbcEntity.setAttribute("qualifiedName", getRdbmsQualifiedName(databaseName, tableName))
-    jdbcEntity.setAttribute("name", tableName)
-
-    AtlasEntityWithDependencies(jdbcEntity)
-  }
-
-  /**
-   * Constructs the the full qualified name of the databse
-   *
-   * @param databaseName
-   * @param tableName
-   * @return
-   */
-  private def getRdbmsQualifiedName(databaseName: String, tableName: String): String =
-    s"${databaseName.toLowerCase}.${tableName.toLowerCase}"
-
+  // scalastyle:off
+  // This is based on the logic how Hive Hook defines qualifiedName for Hive storage desc (borrowed from Apache Atlas v1.1)
+  // https://github.com/apache/atlas/blob/d1f763ab2693911e564a91de29e5c4f2b5976ab7/addons/hive-bridge/src/main/java/org/apache/atlas/hive/bridge/HiveMetaStoreBridge.java#L883-L885
+  // NOTE: there's a chance this method can't guarantee to provider same qualified name - as it relies on "random".
+  // That said, to ensure correctness, temporary table cannot be supported here.
+  // scalastyle:on
   def hiveStorageDescUniqueAttribute(
       cluster: String,
       db: String,
@@ -269,17 +270,27 @@ object external {
     AtlasEntityWithDependencies(sdEntity)
   }
 
+  // scalastyle:off
+  // This is based on the logic how Hive Hook defines qualifiedName for Hive table (borrowed from Apache Atlas v1.1)
+  // https://github.com/apache/atlas/blob/d1f763ab2693911e564a91de29e5c4f2b5976ab7/addons/hive-bridge/src/main/java/org/apache/atlas/hive/bridge/HiveMetaStoreBridge.java#L843-L863
+  // NOTE: there's a chance this method can't guarantee to provider same qualified name - as it relies on "random".
+  // That said, to ensure correctness, temporary table cannot be supported here. (Here I will comment it out.)
+  // scalastyle:on
   def hiveTableUniqueAttribute(
       cluster: String,
       db: String,
       table: String,
       isTemporary: Boolean = false): String = {
     val tableName = if (isTemporary) {
+      /*
       if (SessionState.get() != null && SessionState.get().getSessionId != null) {
         s"${table}_temp-${SessionState.get().getSessionId}"
       } else {
         s"${table}_temp-${RandomStringUtils.random(10)}"
       }
+      */
+      throw new IllegalStateException("Hive temporary table is not supported in SAC!" +
+        s"cluster $cluster / db $db / table $table")
     } else {
       table
     }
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
cc. @aadarshjajodia @ankuriitg for more eyes to review.

@aadarshjajodia
I've found that `hive_storagedesc` entity doesn't show relationship (please take a look at above in snapshots), whereas `hive_table` entity shows relationship with `hive_storagedesc`. The behavior is a bit different from what I was guided - What I understand is, we can set relationship attribute at any side and only one side link is needed.

Should we set attribute value for `table` in `hive_storagedesc` as well for this case? Or direction matters? (like `hive_storagedesc` entity should set relationship to `hive_table` entity, and `hive_table` entity should set relation to `hive_db` entity) Or just a bug?

As you can see from snapshots, we don't see same issue for Spark models. (These models are still provided by SAC side via REST API.)

arunmahadevan:
@HeartSaVioR, a couple of points.

1. IIUC, with this patch SAC creates `hive_` atlas entities even if the catalog is managed by spark in embedded or remote DB, correct? (it only relies on the condition `spark.sql.catalogImplementation == "hive"` which is true even for spark managed catalog).

This has the possibility of un-intended conflicts if un-related tables with same name is created by hive (in HMS) and spark application using spark catalog. Ideally SAC should only create `spark_` entities whereas HMS should be the only one creating `hive_` entities.

So I think unless the catalog is managed by HMS (E.g. if `hive.metastore.uris` is set in hiveConf) SAC should create `spark_` entities, always.

2. If the spark catalog is managed in HMS - with the proposed approach both HMS atlas hook and SAC would create `hive_` entities which is un-necessary. In this case I guess we should just expect HMS to create the `hive_` entities (via HMS atlas hook) and not send duplicate request to Atlas to create the entities.

aadarshjajodia:
@HeartSaVioR with this new relationship attributes model, whenever a relationship is defined as a COMPOSITION, it means the end points in the relationship cannot exist without the other. So for example a hive storage desc cannot exist without a hive table. Another example is a column cannot exist without a table. In such cases as you can see in the json, when we specify the isContainer = false in one of the endpoints of such a relationship that end point must be specified as a relationshipAttribute. So for example in this case, you have to specify the table as a relationship attribute while creating the storage desc entity. I hope this makes it clear

For an AGGREGATION relation(this is something like a database can exist without any tables), in this case the isContainer constraints are not valid like the previous case

@arunmahadevan - I agree. As long as Spark knows that the table has been created in HMS, then SAC should create hive entities. For all other cases it should create spark entities.

For the second point you mentioned, I still feel that if Spark event comes to SAC only when the table has been created in HMS, then Spark need not send duplicate messages to Atlas, since we will get information about these entities from HMS Hook. Even if you send, as long as the **qualifiedName** matches, Atlas will ignore it, but you can avoid sending unnecessary notifications. However if you feel its safer for now, you can go ahead and send these notifications to Atlas. You just need to ensure that the qualifiedName is created in the same way.
Also what should be the behavior then if HMS Hook is disabled. SAC can make a check in the Spark Configs to see if HMS Hook is disabled. If that is the case, only then SAC can send the hive entity messages to Atlas.

HeartSaVioR:
@arunmahadevan @aadarshjajodia

> 1. IIUC, with this patch SAC creates hive_ atlas entities even if the catalog is managed by spark in embedded or remote DB, correct? (it only relies on the condition spark.sql.catalogImplementation == "hive" which is true even for spark managed catalog).

Actually I just kept the approach for determining whether the kind of table is Hive or not - this approach has been used for both SAC and Spark-Navigator. But if we want to define the table is hive only when hive is enabled and remote HMS uris are set, I can make the change. 
(The overall assumption seems that Atlas cluster receives Hive entities from "only one HMS" and "only one catalog", as nothing is being added to the qualifiedName for Hive models.)

Btw, should we avoid classifying hive table via `provider` in table definition then? Or use both? That has been unclear part on me while discussing related thing.

> 2. If the spark catalog is managed in HMS - with the proposed approach both HMS atlas hook and SAC would create hive_ entities which is un-necessary. In this case I guess we should just expect HMS to create the hive_ entities (via HMS atlas hook) and not send duplicate request to Atlas to create the entities.

This is exactly what we've discussed as "possibility of out of order" - SAC doesn't know whether HMS made a change to Atlas (in other words, Atlas processed the events from HMS) prior to events SAC sends out regarding same entity. And if the expectation goes wrong, the request in SAC will silently fail (SAC cannot know whether the request is successful.) and we will have to dig Atlas log to find that request from SAC has been failed. That's the bad part of Kafka hook approach.

HeartSaVioR:
@aadarshjajodia 
> SAC can make a check in the Spark Configs to see if HMS Hook is disabled.

I'm not sure how it can be possible (which Spark config has HMS hook information?), and even it exists it can be changed while spark process is running which SAC has no idea whether the value is changed or not. This is another one of my concern which led me to think always sending creation request is right approach in SAC's point of view though it's redundant.

HeartSaVioR:
@aadarshjajodia 

> So for example in this case, you have to specify the table as a relationship attribute while creating the storage desc entity. I hope this makes it clear

Got it. Btw, we may want to do the same for Spark models, right? Relationship between DB-tables / table-storagedesc / table-columns should be same between Hive models and Spark models.

aadarshjajodia:
> @aadarshjajodia
> 
> > So for example in this case, you have to specify the table as a relationship attribute while creating the storage desc entity. I hope this makes it clear
> 
> 1. This means we don't need to specify storage desc entity as relationship attribute when creating table entity, right? Since it is contradiction that both entities are created prior to each other.
I believe the answer to this is yes. As long as you use relationship attributes and specify the table in the storage desc it should be fine.
> 2. We may want to do the same for Spark models, right? Relationship between DB-tables / table-storagedesc / table-columns should be same between Hive models and Spark models.
Yes that is correct relationship between the spark models should be exactly the same as that present in the hive models. Which brings me to another question, Hive has a column_lineage entity which is used to show the column lineage. Do we not plan to show something like that in the spark models?

HeartSaVioR:
@aadarshjajodia 
HMS is the central one which all DDL events are happening, so easy to ensure the information is in sync with Atlas entity. The same is not applied to SAC - concurrent spark applications will run in multiple nodes and we have to tolerate some missing events or events out of order, hence we decided to stop tracking columns in Spark tables. We may want to track the column later, but not now.

arunmahadevan:
>Btw, should we avoid classifying hive table via provider in table definition then? Or use both? That has been unclear part on me while discussing related thing.

I guess "hive" provider can apply to HMS and spark managed catalog. That being the case we better rely on if a thrift URL is configured (via hive.metastore.uris). WDYT?

>This is exactly what we've discussed as "possibility of out of order" - SAC doesn't know whether HMS made a change to Atlas.

I guess we use that assumption while referring to the entities for Hive Warehouse Connector already? I guess it may be ok to assume that if user has configured HMS and they want to track lineage, they better configure the Atlas hook properly.

Regarding out of order, we need more clarity from Atlas team. Assuming that the events are pushed into the same Kafka partition from both HMS and SAC it should provide ordering. @aadarshjajodia could you confirm?

The duplicate push from Spark side may be ok, however there could be attributes that would apply only to Spark entities (e.g. we added the schema, provider etc) vs Hive entities. So SAC needs to handle it differently.

HeartSaVioR:
> I guess "hive" provider can apply to HMS and spark managed catalog. That being the case we better rely on if a thrift URL is configured (via hive.metastore.uris). WDYT?

It's OK if we are 100% sure tables which provider is not "hive" but remote HMS is set are also managed from remote HMS. I'm not familiar with this so wanted to rely on provider, but if we are sure about the behavior, it totally makes sense.

> I guess we use that assumption while referring to the entities for Hive Warehouse Connector already?

This is not the same case - if my understanding is right, HWC doesn't deal with the query which incurs "create table". It always works with existing table.

https://github.com/hortonworks-spark/spark-atlas-connector/blob/39e47f95fdf3a8d8f9386444361d5a6906fce8cc/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala#L331-L354

```
* Hive Warehouse Connector currently supports four types of operations:
     *   1. SQL / DataFrame Read (batch read)
     *   2. SQL / DataFrame Write (batch write)
     *   3. SQL / DataFrame Write in streaming manner (batch write in streaming)
     *   4. Structured Streaming Write (streaming write)
```

HeartSaVioR:
@arunmahadevan @aadarshjajodia 
I just found the case - suppose HMS hook was disabled at some times. SAC has no idea which tables were created at that moment, and relying on Atlas entities when HMS hook was disabled will all fail (silently in point of SAC's view, and Atlas log only has these error messages). It is not related to whether HMS hook is enabled or disabled now.

aadarshjajodia:
> @aadarshjajodia
> I just found the case - suppose HMS hook was disabled at some times. SAC has no idea which tables were created at that moment, and relying on Atlas entities when HMS hook was disabled will all fail (silently in point of SAC's view, and Atlas log only has these error messages). It is not related to whether HMS hook is enabled or disabled now, so checking whether HMS hook is enabled or not doesn't make sense.
> cc. @arunmahadevan

Okay we can go ahead with the approach of SAC publishing the hive entities as well and not replying on HMS. We just need to make sure the qualifiedName is correct.

arunmahadevan:
@HeartSaVioR @aadarshjajodia If we are for pushing Hive entities from SAC, dont we want to make sure that SAC is pushing the entities conforming to the latest Hive models? We cannot assume spark model is same as hive and push the spark entity as hive entity to Atlas correct? If the attributes are missing or extra attributes are present in the entity wont the entity creation / update fail? I think its going to be a pain to keep updating the SAC code whenever there are any changes to Hive model.

HeartSaVioR:
@arunmahadevan 
Most of attributes are defined as `optional` - both in Hive models and Spark models. I guess create request works like `upsert` and it will not touch existing attributes if there're no attributes specified in creation request. 

And in order to keep compatibility with existing entities, Hive models have to be changed for adding `optional` attributes if any instead of adding `required`, which would be OK for SAC side as well.

@aadarshjajodia Is my expectation correct?

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -182,16 +182,41 @@ object external {
...
-  // In Atlas, the attributes which are not in definition are ignored with WARN messages.
+  // Note: SAC always creates Hive table and corresponding entities, whereas SAC refers
+  // existing table entity on HWC.
 
-  val HIVE_DB_TYPE_STRING = metadata.DB_TYPE_STRING
-  val HIVE_STORAGEDESC_TYPE_STRING = metadata.STORAGEDESC_TYPE_STRING
-  val HIVE_TABLE_TYPE_STRING = metadata.TABLE_TYPE_STRING
+  val HIVE_DB_TYPE_STRING = "hive_db"
+  val HIVE_STORAGEDESC_TYPE_STRING = "hive_storagedesc"
+  val HIVE_TABLE_TYPE_STRING = "hive_table"
```
arunmahadevan: Will this work unless the Hive Entity we create from SAC confirms to the model defined by Hive ? I guess we need to keep the SAC logic in sync with Hive models else things can break. (E.g. if hive decides to change their model). Should we rely only on HMS to create these entities since they are the ones maintaining the Hive Atlas model as well?

HeartSaVioR: Just commented regarding this.

arunmahadevan: If we can expect Hive to always maintain backward compatibility and the model works like upsert may be OK.

ankuriitg:
> I think its going to be a pain to keep updating the SAC code whenever there are any changes to Hive model.

Sorry for the late chime in. I agree with the overall approach that we should create hive entity and send that to Kafka and not worry about whether Hive Hook is enabled or not. SAC should not need to worry about that.

Regarding @arunmahadevan point, I agree that it is going to be pain to make sure we are compatible with the latest Hive models but I think we will have to take that pain as spark is creating Hive tables. I think same will be true for Impala as well. Additionally, I think that pain will not be high as models are not supposed to change, only attributes should be added and they should be optional.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -182,16 +182,41 @@ object external {
...
 
+  // scalastyle:off
+  // This is based on the logic how Hive Hook defines qualifiedName for Hive DB (borrowed from Apache Atlas v1.1)
```
ankuriitg: Can you change these to regular javadoc comments?

arunmahadevan:
>spark is creating Hive tables. I think same will be true for Impala as well.

Since spark is sending request to HMS which is creating the tables, it makes sense for HMS to sent the hive_ events to Atlas. I thought impala does not additionally send the hive events to atlas and relies only on HMS hook. cc @aadarshjajodia.

aadarshjajodia:
> > spark is creating Hive tables. I think same will be true for Impala as well.
> 
> Since spark is sending request to HMS which is creating the tables, it makes sense for HMS to sent the hive_ events to Atlas. I thought impala does not additionally send the hive events to atlas and relies only on HMS hook. cc @aadarshjajodia.

Yes @ankuriitg we had a discussion yesterday with @HyukjinKwon and @arunmahadevan. SAC will not send any hive events. It will depend on HMS hook to send the hive events just like Impala.

HyukjinKwon:
Seems like there was miscommunication. I wasn't there - maybe Jungtaek :D. Yes, if the codes have not changed, SAC does not send any Hive events.

HeartSaVioR:
Just edited mention in comment. Thanks for noticing @HyukjinKwon !

HeartSaVioR:
#245 is the new patch which includes all the changes regarding recent decisions. Some of decision is still taking place so the PR is tagged as WIP for now. Please take a look at #245 instead.

arunmahadevan:
Can you close this then as we would be pursuing #245 ?

HeartSaVioR:
Sure. Let's close this.

PR #247, opened 2019-05-16
[SAC-246][CORE] Use current db name in current Spark session when Spark table information doesn't provide db name

### LINKED ISSUE #246: Current database should be used when Spark table information doesn't provide db name
If we call USE <db> and execute queries without specifying db explicitly, objects will be bound to such db, but tableIdentifier.database is `None`. SAC just provides "default" when it's `None` which makes table entity being matched to wrong database.

HeartSaVioR:

This patch proposes to use current DB name in current Spark session instead of `default` when Spark table information doesn't contain database name.

## How was this patch tested?

Modified UT. Without the change, `prepareEntity` fails to find table entity from "default" database.

Closes #246

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
@@ -201,8 +201,8 @@ object CommandsHarvester extends AtlasEntityUtils with Logging {
   }
 
   def prepareEntity(tableIdentifier: TableIdentifier): AtlasEntityWithDependencies = {
-    val tableName = tableIdentifier.table
-    val dbName = tableIdentifier.database.getOrElse(SparkUtils.getCurrentDatabase)
+    val tableName = SparkUtils.getTableName(tableIdentifier)
+    val dbName = SparkUtils.getDatabaseName(tableIdentifier)
     val tableDef = SparkUtils.getExternalCatalog().getTable(dbName, tableName)
     tableToEntity(tableDef)
   }
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
@@ -202,7 +202,7 @@ object CommandsHarvester extends AtlasEntityUtils with Logging {
 
   def prepareEntity(tableIdentifier: TableIdentifier): AtlasEntityWithDependencies = {
     val tableName = tableIdentifier.table
-    val dbName = tableIdentifier.database.getOrElse("default")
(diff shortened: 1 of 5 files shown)


spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
```diff
@@ -202,7 +202,7 @@ object CommandsHarvester extends AtlasEntityUtils with Logging {
 
   def prepareEntity(tableIdentifier: TableIdentifier): AtlasEntityWithDependencies = {
     val tableName = tableIdentifier.table
-    val dbName = tableIdentifier.database.getOrElse("default")
+    val dbName = tableIdentifier.database.getOrElse(SparkUtils.getCurrentDatabase)
```
arunmahadevan: Do we also need to worry about the case sensitivity? In the spark code I see it converts the DB name to lowercase based on the setting ([here](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala#L130)). Just wondering if something similar needs to be done.

HeartSaVioR: Good point. Spark also has similar code for what we're doing here, so we could follow with it:

https://github.com/apache/spark/blob/6a317c8f014557dfd60931bd1eac3f545520d939/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala#L403-L405

In fact, `currentDb` should have applied `formatDatabaseName` before assigning, so effectively we can just use it,

https://github.com/apache/spark/blob/6a317c8f014557dfd60931bd1eac3f545520d939/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala#L254-L267

but Spark calls `formatDatabaseName` anyway so ideal to follow their pattern.

ankuriitg: You are still planning to make this change for both tableName and dbName, right? And I am guessing you may have already made sure that everywhere we are using SparkUtils methods to find dbName and tableName, but I just wanted to call that out.

arunmahadevan:
Thanks for addressing name formatting. LGTM.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/CommandsHarvester.scala
HeartSaVioR: Yes I found all accessing `tableIdentifier` and changed them via 1c070f3.

HeartSaVioR: I found this change is not outdated while talking with @ankuriitg - sorry I'll make the change quickly.

HeartSaVioR:
Please take a look at latest commit as well. Thanks!

PR #249, opened 2019-05-17
[SAC-248][CORE] Drop attribute `executionTime` in `spark_process`

### LINKED ISSUE #248: Drop attribute `executionTime` in `spark_process` model
`executionTime` attribute in `spark_process` actually represents elapsed time on query execution, whereas `spark_process` is assigned per application. The value doesn't make sense whenever the application runs multiple queries and could mislead end users. We would be better to just remove the attribute.

HeartSaVioR:

This patch proposes dropping `executionTime` attribute as `executionTime` attribute actually simply doesn't make sense for `spark_process` - `spark_process` is assigned per application, whereas Spark provides the value per query.

## How was this patch tested?

UTs. The code change is obvious so didn't do manual tests.

This closes #248

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessor.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessor.scala
@@ -17,9 +17,10 @@
 
 package com.hortonworks.spark.atlas.sql
 
+import java.util.concurrent.TimeUnit
+
 import com.hortonworks.spark.atlas.sql.CommandsHarvester.WriteToDataSourceV2Harvester
 import com.hortonworks.spark.atlas.sql.SparkExecutionPlanProcessor.SinkDataSourceWriter
-
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.command._
 import org.apache.spark.sql.execution.datasources.{InsertIntoHadoopFsRelationCommand, SaveIntoDataSourceCommand}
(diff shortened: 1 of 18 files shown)


HeartSaVioR:
cc. @aadarshjajodia 
This is the change regarding my recent review comment in https://reviews.apache.org/r/70600/ - I feel the reason why you mislead the type of this attribute is the name, so renaming the attribute to make clear the intention.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessor.scala
```diff
@@ -17,25 +17,44 @@
...
+    sink: Option[SinkProgress] = None)
+
+object QueryDetail {
+  def fromQueryExecutionListener(qe: QueryExecution, durationNs: Long): QueryDetail = {
+    val durationMs = TimeUnit.MILLISECONDS.convert(durationNs, TimeUnit.NANOSECONDS)
+    QueryDetail(qe, AtlasUtils.issueExecutionId(), durationMs, Option(SQLQuery.get()))
+  }
+
+  def fromStreamingQueryListener(qe: StreamExecution, event: QueryProgressEvent): QueryDetail = {
+    QueryDetail(qe.lastExecution, AtlasUtils.issueExecutionId(), -1, None,
```
arunmahadevan: Why is it -1 here ? StreamingQueryProgress has a `durationMs` field which gives duration for different phases of the SS query that we could use.

HeartSaVioR: That is per micro-batch, which doesn't fit to QueryDetail.

Btw, as we assign `spark_process` entity per application, the attribute actually simply doesn't make sense. It only makes sense when `spark_process` is assigned per query. Before we change the unit of `spark_process`, we might be able to remove the attribute or rename & replace with application start timestamp if possible, but let's file a new issue to do it.

arunmahadevan: The duration updated for SQL query (via fromQueryExecutionListener) is also per query correct ?  So the same thing applies there too (e.g. there could be more than one query in an application) ?

HeartSaVioR: Yes maybe we can just drop the attribute until we change the unit of `spark_process` as query based one. Also discussed about this with @ankuriitg regarding this. If you also agree I'll make the change here.

Issue #253, opened 2019-05-27
Streaming query should evaluate all events from streaming listener, but should update efficiently


HeartSaVioR:
Currently, we only evaluate streaming event only once for each query runId. This is like a "must-have" thing since otherwise we end up with evaluating plans and updating to Atlas every time.

But I've also found the case where the micro-batch is not pulling data from entire sources (e.g. the query is configured to pull from 3 topics but the specific batch only pulls from 1 topics which ends up exposing one topics in physical plan).

So unfortunately, we should evaluate all events to have aggregated sources (which is actually not guaranteed to be whole entity), but update the entities efficiently. (Likewise only updating when there's any update in inputs or outputs.)

HeartSaVioR:
Other approach is also leveraging logical plan, but need to be really careful not to bring regression.

PR #254, opened 2019-05-27
[SAC-252][CORE] Skip evaluating streaming event when query is not physically executed yet

### LINKED ISSUE #252: NPE occurred in SAC side when executing Streaming query which there's no new input data
Screen Shot 2019-05-27 at 4 03 09 PM

This occurs consistently with the case: executing streaming query which checkpoint is available, but new input data is not ready to pull, hence no new batch is started from Spark. In this case, streaming listener could still report status before any micro-batch is executed which the last execution of physical plan is still not prepared.

HeartSaVioR:

Please refer #252 to see the rationalization of this patch.

This patch fixes the intermittent NPE occurred in SAC side when executing Streaming query, which is occurred due to receiving streaming event before executing actual batch. This can be easily reproduced via executing streaming query and let query consuming all the available input data, and stop the query, and restart the query in new spark session (spark-shell) which leverage same checkpoint.

This patch checks whether the last executed physical plan is available, and skips processing it if the plan is not available. This effectively makes SAC to wait until the plan is available (since streaming event will be sent at least per each batch, so eventually it will be available).

## How was this patch tested?

Manually tested: patched SAC didn't throw NPE with no batch but just waited for new event, and when I provided new records to one of topic, SAC picked the event and created a new lineage.
 
Screen Shot 2019-05-27 at 5 23 07 PM

NOTE: there're actually three topics in source side, but only one source is captured - it's a known-issue and I filed #253 to track this.

This closes #252

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/SparkAtlasStreamingQueryEventTracker.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/SparkAtlasStreamingQueryEventTracker.scala
@@ -53,18 +53,26 @@ class SparkAtlasStreamingQueryEventTracker(
     if (!streamQueryHashset.contains(event.progress.runId)) {
       val query = SparkSession.active.streams.get(event.progress.id)
       if (query != null) {
-        query match {
+        val qd = query match {
           case query: StreamingQueryWrapper =>
-            val qd = QueryDetail.fromStreamingQueryListener(query.streamingQuery, event)
-            executionPlanTracker.pushEvent(qd)
-            streamQueryHashset.add(event.progress.runId)
+            Some(QueryDetail.fromStreamingQueryListener(query.streamingQuery, event))
 
           case query: StreamExecution =>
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
> So SAC will create a new lineage if the query is restarted? Is that because its tied to spark application id and the application id changes during restart?

Yes, and I have no idea why the decision has been made. One suspect is that events from catalog don't have query id, but there's other value we can get when spark_process tracks "application", so not sure. Interestingly, Spark-Navigator did same decision, which I would feel safer to stick the policy, unless we experiment changing this and validate the change doesn't incur regression or conceptual limitation.

arunmahadevan:
>One suspect is that events from catalog don't have query id

The `QueryStartedEvent` in the `StreamingQueryListener` provides query-id and query-name. For batch job perhaps the `SparkListenerJobStart.properties()` provides some detail but I agree we need more investigation to change the approach

Anyways I feel the current approach does not provide much value for periodically scheduled batch jobs and for SS queries when the queries are restarted. Created https://github.com/hortonworks-spark/spark-atlas-connector/issues/261 to track it.

PR #255, opened 2019-05-28
[SAC-253][SQL] Streaming query should evaluate all events from streaming listener, but should update efficiently

### LINKED ISSUE #253: Streaming query should evaluate all events from streaming listener, but should update efficiently

HeartSaVioR:

Please refer #253 to see rationalization of this patch.

This patch proposes to evaluate all streaming events from streaming listener to capture late activated sources and sinks in streaming query. This change will incur huge requests to Atlas (at least one per batch), so this patch also introduces some optimization on sending request to Atlas.

The approach of optimization is memorizing accumulated inputs and outputs per streaming query and only sending request when either new inputs or new outputs is not a subset of accumulated inputs or outputs.

## How was this patch tested?

Added UTs

* UTs for class doing optimization
* another UT for simulating late source activation on streaming query

(skipping manual tests since one of new UT plays like acceptance test.)

This closes #253

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/AtlasEntityCreationRequestHelper.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/AtlasEntityCreationRequestHelper.scala
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hortonworks.spark.atlas
+
+import java.util.UUID
+
+import com.hortonworks.spark.atlas.types.metadata
+import com.hortonworks.spark.atlas.utils.Logging
+import org.apache.atlas.model.instance.AtlasObjectId
+
+import scala.collection.mutable
+
+class AtlasEntityCreationRequestHelper(
+    atlasClient: AtlasClient) extends Logging {
+  // query to (inputs, outputs)
+  private val queryToInputsAndOutputs = new mutable.HashMap[UUID,
+    (Set[AtlasObjectId], Set[AtlasObjectId])]()
+
+  def requestCreation(entities: Seq[SACAtlasReferenceable], runId: Option[UUID] = None): Unit = {
+    runId match {
+      case Some(rid) => updateEntitiesForStreamingQuery(rid, entities)
+      case None => updateEntitiesForBatchQuery(entities)
+    }
+  }
+
+  private def updateEntitiesForBatchQuery(entities: Seq[SACAtlasReferenceable]): Unit = {
+    // the query is batch, hence always create entities
+    // create input/output entities as well as update process entity(-ies)
+    createEntities(entities)
+  }
+
+  private def updateEntitiesForStreamingQuery(
+      runId: UUID,
+      entities: Seq[SACAtlasReferenceable]): Unit = {
+    // the query is streaming, so which partial of source/sink entities can be seen
+    // in specific batch - need to accumulate efficiently
+    val processes = entities
+      .filter(en => en.typeName == metadata.PROCESS_TYPE_STRING
+        && en.isInstanceOf[SACAtlasEntityWithDependencies])
+      .map(_.asInstanceOf[SACAtlasEntityWithDependencies])
+
+    val inputs = processes.flatMap { p =>
+      AtlasEntityReadHelper.getSeqAtlasObjectIdAttribute(p.entity, "inputs")
+    }.toSet
+
+    val outputs = processes.flatMap { p =>
+      AtlasEntityReadHelper.getSeqAtlasObjectIdAttribute(p.entity, "outputs")
+    }.toSet
+
+    queryToInputsAndOutputs.get(runId) match {
+      case Some((is, os)) if !inputs.subsetOf(is) || !outputs.subsetOf(os) =>
+        // The query is streaming, and at least either inputs or outputs is not a
+        // subset of accumulated one.
+
+        // NOTE: we leverage the 'process' model's definition:
+        // inputs and outputs are defined as set in definition, and Atlas automatically
+        // accumulate these values which doesn't require us to track all inputs and
+        // outputs and always provide accumulated one.
+        // If we need to do in our own, we should also accumulate inputs and outputs
+        // in SparkCatalogEventProcessor and maintain full of inputs and outputs.
+        // Here we only accumulate inputs/outputs for each streaming query (runId).
+
+        createEntities(entities)
+
+        // update inputs and outputs as accumulating current one and new inputs/outputs
+        updateInputsAndOutputs(runId, is.union(inputs), os.union(outputs))
+
+      case Some((_, _)) => // if inputs.subsetOf(is) && outputs.subsetOf(os)
+      // we already updated superset of inputs/outputs, skip updating
+
+      case _ =>
+        // the streaming query hasn't been examined in current session
+        createEntities(entities)
+
+        // update inputs and outputs as new inputs/outputs, as there's nothing to accumulate
+        updateInputsAndOutputs(runId, inputs, outputs)
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/SparkAtlasStreamingQueryEventTracker.scala
```diff
@@ -39,8 +39,6 @@ class SparkAtlasStreamingQueryEventTracker(
     this(new AtlasClientConf)
   }
 
-  private val streamQueryHashset = new mutable.HashSet[java.util.UUID]
```
HeartSaVioR: This is the key point of patch's changeset: we no longer evaluate streaming event only once, which would miss lately activated sources and sinks.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessor.scala
```diff
@@ -145,19 +153,18 @@ class SparkExecutionPlanProcessor(
         //        if (r.relation.getClass.getCanonicalName.endsWith("dd")) =>
         //      println("close hive connection via " + r.relation.getClass.getCanonicalName)
 
-      } ++ {
```
HeartSaVioR: Just fixed the indentation.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/internal.scala
```diff
@@ -179,9 +179,9 @@ object internal extends Logging {
   }
 
   def etlProcessToEntity(
-                          inputs: Seq[SACAtlasReferenceable],
```
HeartSaVioR: Just fixed the indentation here and below.

spark-atlas-connector/src/test/scala/com/hortonworks/spark/atlas/sql/SparkExecutionPlanProcessorForStreamingQuerySuite.scala
```diff
@@ -183,6 +183,72 @@ class SparkExecutionPlanProcessorForStreamingQuerySuite
     }
   }
 
+  test("SAC-253 Kafka source(s) to kafka sink - micro-batch query" +
```
HeartSaVioR: This UT is basically running same query with above UT, but activating 1 topic per batch via sending message to topic later.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/SparkAtlasStreamingQueryEventTracker.scala
```diff
@@ -50,30 +48,25 @@ class SparkAtlasStreamingQueryEventTracker(
...
     logInfo(s"Track running Spark Streaming query in the Spark Atlas: $event")
-    if (!streamQueryHashset.contains(event.progress.runId)) {
-      val query = SparkSession.active.streams.get(event.progress.id)
-      if (query != null) {
-        query match {
-          case query: StreamingQueryWrapper =>
-            val qd = QueryDetail.fromStreamingQueryListener(query.streamingQuery, event)
-            executionPlanTracker.pushEvent(qd)
-            streamQueryHashset.add(event.progress.runId)
+    val query = SparkSession.active.streams.get(event.progress.id)
```
arunmahadevan: Can we relay StreamingQueryProgress.sources (i.e. the source progress) to decide whether to process the event ? i.e if the number of sources changed or the topic/partition data in the offset changed (for Kafka) ? Its tricky but thinking if it can avoid parsing the plan each time.

HeartSaVioR: I agree it would be efficient, but I don't prefer relying on string parse which we should do for extracting informations from `StreamingQueryProgress.sources` - for sink I had to do this because DSv1 sink doesn't have its logical plan.

I'm OK to apply this if we could just blindly compare the "sets" of source descriptions (this means we get rid of numbers which would be changed for each batch) between batches. I guess it would be incomplete, so put this in first level of check, and keep the newly added check as second level of check. What do you think?

HeartSaVioR: I just realized my idea cannot be applied, since descriptions in `StreamingQueryProcess.sources` will provide same even only part of sources are activated in this batch.

This is the set of descriptions in `StreamingQueryProcess.sources` when the micro-batch only activates one source, `sparkread0`:

```
sources

0 = "KafkaV2[SubscribePattern[sparkread0[1-3]]]"
1 = "KafkaV2[Subscribe[sparkread04, sparkread05]]"
2 = "KafkaV2[Assign[sparkread06-0, sparkread06-1, sparkread06-2, sparkread06-3, sparkread06-4, sparkread06-5, sparkread06-6, sparkread06-7, sparkread06-8, sparkread06-9, sparkread07-0, sparkread07-1, sparkread07-2, sparkread07-3, sparkread07-4, sparkread07-5, sparkread07-6, sparkread07-7, sparkread07-8, sparkread07-9]]"
```

So we should use same approach for checking whether we can skip processing or not, and populating sources.

One thing you may also notice from the above string representation - we can't get exact topic list from subscribe pattern if we deal with logical plan or `StreamingQueryProcess.sources`. (`[1-3]` is easy pattern but you may want to provide `*` or `?` or whatever which we can't populate all possible topics.) 

Hopefully, SAC deals with executed physical plan which contains actual Input(RDD)Partition to read for batch/streaming query - which is closer to track actual impacts which query is introducing. We may not want to lose this functionality.

HeartSaVioR:
@arunmahadevan 
Could you take another look at my comment? There's pros and cons about which thing (streaming progress, logical plan, physical plan) we leverage to populate inputs and outputs, and I'd like to keep leveraging physical plan for DSv1/DSv2 writer (DF query will fall into this case).

It would be awesome if you could finalize your review in this week, so that we can at least merge this in early next week. Thanks in advance!

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/SparkAtlasStreamingQueryEventTracker.scala
arunmahadevan: I think you have to look at the startOffset and endOffset (inside SourceProgress) which will have the expanded topic and list of partitions. Yes its kind of tricky and maybe it will work only for Kafka. (e.g. If there are some other sources that stores offsets differently we have to invent a new way). If its not straightforward, ok to parse the plan.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/AtlasEntityCreationRequestHelper.scala
```diff
@@ -0,0 +1,107 @@
...
+
+    val inputs = processes.flatMap { p =>
+      AtlasEntityReadHelper.getSeqAtlasObjectIdAttribute(p.entity, "inputs")
+    }.toSet
+
+    val outputs = processes.flatMap { p =>
+      AtlasEntityReadHelper.getSeqAtlasObjectIdAttribute(p.entity, "outputs")
+    }.toSet
+
+    queryToInputsAndOutputs.get(runId) match {
```
arunmahadevan: The `runId` can change if a query is stopped and re-started within same spark application. Since we want to track the lineage based on application Id, we don't want to re-create the lineage if a query is just restarted within same application correct? If so the key should be query-id here.

HeartSaVioR: Nice finding. Will address shortly.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/SparkAtlasStreamingQueryEventTracker.scala
HeartSaVioR: @arunmahadevan 

> I think you have to look at the startOffset and endOffset (inside SourceProgress) which will have the expanded topic and list of partitions.

Yeah thanks for the info. I've missed this. I'd say that we rely on too many parts (logical plan, physical plan, and sink for streaming query status) and adding more sounds to have more chance to be broken, but the approach looks better to determine whether the plan should be re-examined or not.

Given you have idea here, would I ask you to find the time to see whether this approach could be applied to some other sources as well? `file` and `kafka` would be enough if we use that as earlier check and don't get rid of current check. Thanks in advance!

HeartSaVioR:
I'll merge this since I've resolved one comment and left another one to be beyond this PR. If we would like to address the comment, we can do it with new issue/PR.

PR #257, opened 2019-05-29
[SAC-256][BUILD] Exclude Hadoop dependencies from Beeju

### LINKED ISSUE #256: Exclude Hadoop dependencies from Beeju
I've seen version conflict between Spark's Hadoop dependency and Beeju's Hadoop dependency. (This is actually observed when we change the version of Spark.)

We might have to get rid of Beeju if Spark and Beeju are being conflicted and we are forced to choose one. In other words, to leverage Beeju, tests should run correctly even if we exclude Hadoop dependencies from Beeju.

HeartSaVioR:

This patch proposes to exclude Hadoop dependencies from Beeju and leverages Hadoop dependencies from Spark. (Beeju is in "test" scope, which won't make problem in usage.)

The base ground of this decision is, if Hadoop version conflict occurs between Beeju and Spark, we should rely on Spark which cannot be replaceable.

## How was this patch tested?

Existing UTs.

Closes #256

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -274,6 +274,12 @@
       <artifactId>beeju</artifactId>
       <version>1.3.1</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
Merging this since this is a simple change and only affects testing.

Issue #261, opened 2019-05-31
Investigate Lineage tracking based on query / data frame / SQL rather than application id.


arunmahadevan:
SAC currently tracks lineage based on Spark application (appid). This creates new lineage each time a SS query is restarted and even for periodic execution of the same batch job. I guess it would be better to track the lineage by referring the SS query ID or query Name and some attribute of the data frame / SQL for batch queries.

See also https://github.com/hortonworks-spark/spark-atlas-connector/pull/254 for relevant discussions.

PR #262, opened 2019-06-04
[SAC-223][DOC] Document "known limitations", and recent change on Spark models vs Hive models

### LINKED ISSUE #223: Document "Known limitations"
There're several "known limitations" which we decide not to bring the features due to some technical reasons. It would be better to maintain the list to avoid confusion.

HeartSaVioR:

This patch adds explanation on "known limitations" (based on decision decision or other technical reason), as well as recent change on Spark models vs Hive models which is pretty hard to understand unless end users track the code changes.

## How was this patch tested?

N/A

This closes #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,9 @@ This doesn't apply to Hive models, which central remote HMS takes care of DDLs a
 
 > SAC doesn't track dropping tables (Spark models).
 
-Same reason as above. If SAC doesn't support some operations on DDL, it would be basically same reason.
+"drop table" event from Spark only provides db and table name, which is NOT sufficient to create qualifierName - especially we separate two types of tables - spark and hive. 
+
+SAC depends on reading the Spark Catalog to get table information but Spark might already dropped the table when SAC notices the table is dropped, so that is NOT consistent. Due to this technical limitation, we have to drop this feature.
 
 > ML entities/events may not be tracked properly.
 
--- a/README.md
+++ b/README.md
@@ -136,9 +136,9 @@ SAC relies on query listener to retrieve query and examine the impacts.
 
 > All "inputs" and "outputs" in multiple queries are accumulated into single "spark_process" entity when there're multple queries running in single Spark session.
 
-There're pros and cons regarding which is used for unit of "spark_process". Currently SAC uses "application ID" as unit - this approach could concentrate overall impacts on the single Spark process for who ran for when.
+"spark_process" maps to an "applicationId" in Spark. This is helpful as it allows admin to track all changes that occurred as part of an application. But it also causes lineage/relationship graph in "spark_process" to be complicated and less meaningful.
 
-But we also received other voices on this approach - lineage/relationship graph in "spark_process" are too complicated and confused and less meaningful. It's valid voice, we've filed #261 to investigate changing the unit of "spark_process" entity to query. It doesn't mean we will change it soon. It will be address only if we see the clear benefits on changing it.
+We've filed #261 to investigate changing the unit of "spark_process" entity to query. It doesn't mean we will change it soon. It will be addressed only if we see clear benefits of changing it.
 
 > Only part of inputs are tracked in Streaming query.
 
@@ -150,17 +150,17 @@ If your query are running long enough that it ingests data from all topics, it w
 
 > SAC doesn't support tracking changes on columns (Spark models).
 
-We found difficulty to make column entity be consistently "up-to-date" supposing multiple spark applications are running and each SAC in spark application are trying to update on Atlas. Due to the difficulty, we dropped supporting column tracking.
-
-We haven't found difficulty on table level, but once we collect difficulties on making table entity being consistent as well, we might also consider drop supporting DDL and blindly create the entities whenever they're referenced.
+We are investigating how to add support for column entity. The main issue we face is how to make this change consistent when multiple spark applications make changes to the same table/column.
 
 This doesn't apply to Hive models, which central remote HMS takes care of DDLs and Hive Atlas Hook will take care of updates.
 
 > SAC doesn't track dropping tables (Spark models).
 
-"drop table" event from Spark only provides db and table name, which is NOT sufficient to create qualifierName - especially we separate two types of tables - spark and hive. 
+"drop table" event from Spark only provides db and table name, which is NOT sufficient to create qualifierName - especially we separate two types of tables - spark and hive.
 
-SAC depends on reading the Spark Catalog to get table information but Spark might already dropped the table when SAC notices the table is dropped, so that is NOT consistent. Due to this technical limitation, we have to drop this feature.
+SAC depends on reading the Spark Catalog to get table information but Spark might already dropped the table when SAC notices the table is dropped, so that is NOT consistent.
+
+We are investigating how to change Spark to provide necessary information via listener, maybe snapshot of information before deletion happens.
 
 > ML entities/events may not be tracked properly.
 
--- a/README.md
+++ b/README.md
@@ -116,16 +116,16 @@ Spark models vs Hive models
 
 SAC classifies table related entities with two different kind of models: Spark / Hive.
 
-We decided to determine which models SAC should use via taking realistic condition: whether Hive Hook can take care of table related entities or not. For creating Hive entities, SAC is not the best one to fill out information on entities - so we let Hive to do its own.
+We decided to determine which models SAC should create based on whether Hive Hook can take care of table related entities or not. For Hive entities, SAC relies on HMS Atlas hook as the source of truth and expects HMS atlas hook to create it.
 
 As SAC cannot check whether the condition is properly made, SAC relies on precondition: if Spark session is connected to the "remote HMS" (via thrift), SAC assumes Hive Hook is already set up in remote HMS and Hive Hook will take care of entities creation.
 
-In other words, SAC assumes table entities are being created in Hive side and just refers these entities via object id:
+In other words, SAC assumes table entities are being created in Hive side and just refers these entities via object id if below conditions are true:
 
-* The value of "spark.sql.catalogImplementation" is set to "hive"
+* SparkSession.builder.enableHiveSupport is set
 * The value of "hive.metastore.uris" is set to non-empty
 
-For other cases, SAC will create necessary table related entities as Spark models (even it uses Hive metastore).
+For other cases, SAC will create necessary table related entities as Spark models.
 
 One exceptional case is HWC - for HWC source and/or sink, SAC will not create table related entities and always refer to Hive table entities via object id.
 
@@ -138,13 +138,13 @@ SAC relies on query listener to retrieve query and examine the impacts. We may n
 
 > All "inputs" and "outputs" in multiple queries are accumulated into single "spark_process" entity when there're multple queries running in single Spark session.
 
-Unfortunately, we got lost on history for this. We assume this approach could concentrate overall impacts on the single Spark process for who ran for when, but it's not pretty clear why application id was taken.
+There're pros and cons regarding which is used for unit of "spark_process". Currently SAC uses "application ID" as unit - this approach could concentrate overall impacts on the single Spark process for who ran for when.
 
-We've filed #261 to investigate changing the unit of "spark_process" entity to query, but it doesn't mean we will change it soon. It will be address only if we see the clear benefits on changing it.
+But we also received other voices on this approach - lineage/relationship graph in "spark_process" are too complicated and confused and less meaningful. It's valid voice, we've filed #261 to investigate changing the unit of "spark_process" entity to query. It doesn't mean we will change it soon. It will be address only if we see the clear benefits on changing it.
 
 > Only part of inputs are tracked in Streaming query.
 
-This is from design choice on "trade-off": Kafka source supports subscribing with "pattern" which SAC cannot enumerate all matching existing topics, or even all possible topics (even it's possible it doesn't make sense, though). 
+This is from design choice on "trade-off": Kafka source supports subscribing with "pattern" and SAC cannot enumerate all matching existing topics, or even all possible topics (even it's possible it doesn't make sense, though).
 
 We found "executed plan" provides actual topics which each (micro) batch read and processed, but as a result, only inputs which participate on (micro) batch are being included as "inputs" in "spark_process" entity. 
 
@@ -156,7 +156,7 @@ We found difficulty to make column entity be consistently "up-to-date" supposing
 
 We haven't found difficulty on table level, but once we collect difficulties on making table entity being consistent as well, we might also consider drop supporting DDL and blindly create the entities whenever they're referenced.
 
-This doesn't apply to Hive models, which central remote HMS takes care of DDLs and Hive Hook will take care of updates.
+This doesn't apply to Hive models, which central remote HMS takes care of DDLs and Hive Atlas Hook will take care of updates.
 
 > SAC doesn't track dropping tables (Spark models).
 
--- a/README.md
+++ b/README.md
@@ -116,11 +116,9 @@ Spark models vs Hive models
 
 SAC classifies table related entities with two different kind of models: Spark / Hive.
 
-We decided to determine which models SAC should create based on whether Hive Hook can take care of table related entities or not. For Hive entities, SAC relies on HMS Atlas hook as the source of truth and expects HMS atlas hook to create it.
+We decided to skip sending create events for Hive tables managed by HMS to avoid duplication of those events from Atlas hook for Hive . For Hive entities, Atlas relies on Atlas hook for Hive as the source of truth.
 
-As SAC cannot check whether the condition is properly made, SAC relies on precondition: if Spark session is connected to the "remote HMS" (via thrift), SAC assumes Hive Hook is already set up in remote HMS and Hive Hook will take care of entities creation.
-
-In other words, SAC assumes table entities are being created in Hive side and just refers these entities via object id if below conditions are true:
+SAC assumes table entities are being created in Hive side and just refers these entities via object id if below conditions are true:
 
 * SparkSession.builder.enableHiveSupport is set
 * The value of "hive.metastore.uris" is set to non-empty
--- a/README.md
+++ b/README.md
@@ -111,6 +111,64 @@ When running on cluster node, you will also need to distribute this keytab, belo
 
 When Spark application is started, it will transparently track the execution plan of submitted SQL/DF transformations, parse the plan and create related entities in Atlas.
 
+Spark models vs Hive models
+====
+
+SAC classifies table related entities with two different kind of models: Spark / Hive.
+
+We decided to determine which models SAC should use via taking realistic condition: whether Hive Hook can take care of table related entities or not. For creating Hive entities, SAC is not the best one to fill out information on entities - so we let Hive to do its own.
+
+As SAC cannot check whether the condition is properly made, SAC relies on precondition: if Spark session is connected to the "remote HMS" (via thrift), SAC assumes Hive Hook is already set up in remote HMS and Hive Hook will take care of entities creation.
+
+In other words, SAC assumes table entities are being created in Hive side and just refers these entities via object id:
+
+* The value of "spark.sql.catalogImplementation" is set to "hive"
+* The value of "hive.metastore.uris" is set to non-empty
+
+For other cases, SAC will create necessary table related entities as Spark models (even it uses Hive metastore).
+
+One exceptional case is HWC - for HWC source and/or sink, SAC will not create table related entities and always refer to Hive table entities via object id.
+
+Known Limitations (Design decision)
+====
+
+> SAC only supports SQL/DataFrame API (in other words, SAC doesn't support RDD).
+
+SAC relies on query listener to retrieve query and examine the impacts. We may not investigate on supporting RDD for future unless strongly necessary.
+
+> All "inputs" and "outputs" in multiple queries are accumulated into single "spark_process" entity when there're multple queries running in single Spark session.
+
+Unfortunately, we got lost on history for this. We assume this approach could concentrate overall impacts on the single Spark process for who ran for when, but it's not pretty clear why application id was taken.
+
+We've filed #261 to investigate changing the unit of "spark_process" entity to query, but it doesn't mean we will change it soon. It will be address only if we see the clear benefits on changing it.
+
+> Only part of inputs are tracked in Streaming query.
+
+This is from design choice on "trade-off": Kafka source supports subscribing with "pattern" which SAC cannot enumerate all matching existing topics, or even all possible topics (even it's possible it doesn't make sense, though). 
+
+We found "executed plan" provides actual topics which each (micro) batch read and processed, but as a result, only inputs which participate on (micro) batch are being included as "inputs" in "spark_process" entity. 
+
+If your query are running long enough that it ingests data from all topics, it would have all topics in "spark_process" entity.
+
+> SAC doesn't support tracking changes on columns (Spark models).
+
+We found difficulty to make column entity be consistently "up-to-date" supposing multiple spark applications are running and each SAC in spark application are trying to update on Atlas. Due to the difficulty, we dropped supporting column tracking.
+
+We haven't found difficulty on table level, but once we collect difficulties on making table entity being consistent as well, we might also consider drop supporting DDL and blindly create the entities whenever they're referenced.
+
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
Please take a look at the doc addition and comment if there're missing pieces. Ideally I would like to put every known limitations to here so that we don't need to explain individually.

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
 
 When Spark application is started, it will transparently track the execution plan of submitted SQL/DF transformations, parse the plan and create related entities in Atlas.
 
+Spark models vs Hive models
+====
+
+SAC classifies table related entities with two different kind of models: Spark / Hive.
```
arunmahadevan: I am not sure classifies conveys the right meaning since SAC does not create Hive models.
Maybe "SAC creates Spark models when spark manages the catalog and just references Hive meta store created models (Hive models) when the catalog is managed by Hive meta store"

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+We decided to determine which models SAC should use via taking realistic condition: whether Hive Hook can take care of table related entities or not. For creating Hive entities, SAC is not the best one to fill out information on entities - so we let Hive to do its own.
```
arunmahadevan: SAC should use via taking realistic condition -> SAC should create based on ?
For creating Hive entities, SAC is not the best one to fill out information on entities - so we let Hive to do its own -> Somthing like "For Hive entities, SAC relies on HMS Atlas hook as the source of truth and expects HMS atlas hook to create it."

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+As SAC cannot check whether the condition is properly made, SAC relies on precondition: if Spark session is connected to the "remote HMS" (via thrift), SAC assumes Hive Hook is already set up in remote HMS and Hive Hook will take care of entities creation.
```
arunmahadevan: Hive hook -> Hive Atlas hook

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+In other words, SAC assumes table entities are being created in Hive side and just refers these entities via object id:
```
arunmahadevan: via object id if below conditions are true:

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+* The value of "spark.sql.catalogImplementation" is set to "hive"
```
arunmahadevan: spark.sql.catalogImplementation is internal property correct? If so refer to the user exposed property here (SparkSession.builder.enableHiveSupport?)

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+* The value of "hive.metastore.uris" is set to non-empty
+
+For other cases, SAC will create necessary table related entities as Spark models (even it uses Hive metastore).
```
arunmahadevan: `even it uses Hive metastore` kind of contradicts. Maybe just remove this.

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+Known Limitations (Design decision)
+====
+
+> SAC only supports SQL/DataFrame API (in other words, SAC doesn't support RDD).
+
+SAC relies on query listener to retrieve query and examine the impacts. We may not investigate on supporting RDD for future unless strongly necessary.
+
+> All "inputs" and "outputs" in multiple queries are accumulated into single "spark_process" entity when there're multple queries running in single Spark session.
+
+Unfortunately, we got lost on history for this. We assume this approach could concentrate overall impacts on the single Spark process for who ran for when, but it's not pretty clear why application id was taken.
```
arunmahadevan: I am not sure if its ideal to explicitly state `we got lost on history for this` in the README. Maybe just state the current behavior and mention the alternative approach that could be considered in future.

arunmahadevan: Maybe also state the effect of the current approach. (same batch or streaming queries run in a different spark application will create separate lineage)?

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+We've filed #261 to investigate changing the unit of "spark_process" entity to query, but it doesn't mean we will change it soon. It will be address only if we see the clear benefits on changing it.
+
+> Only part of inputs are tracked in Streaming query.
+
+This is from design choice on "trade-off": Kafka source supports subscribing with "pattern" which SAC cannot enumerate all matching existing topics, or even all possible topics (even it's possible it doesn't make sense, though).
```
arunmahadevan: which -> and

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+We found difficulty to make column entity be consistently "up-to-date" supposing multiple spark applications are running and each SAC in spark application are trying to update on Atlas. Due to the difficulty, we dropped supporting column tracking.
+
+We haven't found difficulty on table level, but once we collect difficulties on making table entity being consistent as well, we might also consider drop supporting DDL and blindly create the entities whenever they're referenced.
+
+This doesn't apply to Hive models, which central remote HMS takes care of DDLs and Hive Hook will take care of updates.
+
+> SAC doesn't track dropping tables (Spark models).
+
+"drop table" event from Spark only provides db and table name, which is NOT sufficient to create qualifierName - especially we separate two types of tables - spark and hive.
```
arunmahadevan: Is it meaningful to drop entities? Dont we need the entities to show how the data got transformed (in the same or prior spark sessions) before the table was dropped?

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+> SAC doesn't track dropping tables (Spark models).
+
+"drop table" event from Spark only provides db and table name, which is NOT sufficient to create qualifierName - especially we separate two types of tables - spark and hive. 
+
+SAC depends on reading the Spark Catalog to get table information but Spark might already dropped the table when SAC notices the table is dropped, so that is NOT consistent. Due to this technical limitation, we have to drop this feature.
+
+> ML entities/events may not be tracked properly.
+
+We are concentrating on making basic features be stable: we are not including ML features on the target of basic features as of now. We will revisit once we are sure to resolve most of issues on basic features.
```
arunmahadevan: Do users need to be aware of these details? Just state the current limitation and the proposed patch and the tracking will be improved post Spark 3.0.

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+This is from design choice on "trade-off": Kafka source supports subscribing with "pattern" which SAC cannot enumerate all matching existing topics, or even all possible topics (even it's possible it doesn't make sense, though). 
+
+We found "executed plan" provides actual topics which each (micro) batch read and processed, but as a result, only inputs which participate on (micro) batch are being included as "inputs" in "spark_process" entity. 
+
+If your query are running long enough that it ingests data from all topics, it would have all topics in "spark_process" entity.
+
+> SAC doesn't support tracking changes on columns (Spark models).
+
+We found difficulty to make column entity be consistently "up-to-date" supposing multiple spark applications are running and each SAC in spark application are trying to update on Atlas. Due to the difficulty, we dropped supporting column tracking.
```
arunmahadevan: Maybe just state the behavior ? E.g. `alter tables` are not tracked?

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+We haven't found difficulty on table level, but once we collect difficulties on making table entity being consistent as well, we might also consider drop supporting DDL and blindly create the entities whenever they're referenced.
```
arunmahadevan: Not sure I understand what this means. Is this related to multiple create tables (last one wins ?)

HeartSaVioR:
> Not related but since we are editing the doc, should we also update "Pre-create Atlas models (Recommended)" now that the models are pre-created by Atlas? Maybe completely remove that section or keep it as "optional" step.

Apache Atlas doesn't have any official release which contains spark models. Once we got news on new release, we could update it.

README.md
HeartSaVioR: It is not only related to "create table" but any DDLs which could update table related entities. Actually I'm not sure we really want to support dealing with entities which are belong to "local" (not cluster-wise). We may want to discuss this more, but out of scope for this PR.

README.md
HeartSaVioR: I'm not sure about your point. Are you suggesting not to explain reason and just explain which is not supported?

And we don't even blindly create column entities when table is referenced, even Spark models, so if we state the behavior we also need to specify this as well.

README.md
HeartSaVioR: I intended to add details - this is not only for end users but also for any contributors. Given this is open source, ideally all materials should be here to avoid centralization of knowledge. If this would give confusion, I'll have separate doc to explain the details and leave only overview here.

README.md
HeartSaVioR: If I'm not mistaken, even entity is marked as removed, the information itself is not removed if it's referenced from others. I'm not sure how Atlas team tends to deal with this, but in any way we don't track dropping table so same result.

@aadarshjajodia Could you guide how other hooks deal with dropping entities?

README.md
HeartSaVioR: I agree that's inappropriate statement. Will remove.

Btw, the case end users mostly concerning is "inputs" and "outputs" being accumulated and the overall graph may look ugly. Ideally we need to explain why we take this approach, but we are not clear about this either, so OK to just state it.

> same batch or streaming queries run in a different spark application will create separate lineage

Agreed for streaming queries but don't get it for batch. Can batch run multiple times with same query ID?

README.md
HeartSaVioR: Agreed. GIven we describe the condition here, let's assume end users would not be confused without this additional statement.

README.md
HeartSaVioR: Agree that is internal configuration, we would be better to describe user-facing API. Nice suggestion!

README.md
HeartSaVioR: Much nicer representation. Thanks!

README.md
HeartSaVioR: > SAC creates Spark models when spark manages the catalog and just references Hive meta store created models (Hive models) when the catalog is managed by Hive meta store

This is not true - we still deal with table entities with Spark models if we leverage embedded hive metastore or so. That's why explaining this is not going to be intuitive.

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+We decided to determine which models SAC should create based on whether Hive Hook can take care of table related entities or not. For Hive entities, SAC relies on HMS Atlas hook as the source of truth and expects HMS atlas hook to create it.
+
+As SAC cannot check whether the condition is properly made, SAC relies on precondition: if Spark session is connected to the "remote HMS" (via thrift), SAC assumes Hive Hook is already set up in remote HMS and Hive Hook will take care of entities creation.
+
+In other words, SAC assumes table entities are being created in Hive side and just refers these entities via object id if below conditions are true:
+
+* SparkSession.builder.enableHiveSupport is set
+* The value of "hive.metastore.uris" is set to non-empty
+
+For other cases, SAC will create necessary table related entities as Spark models.
```
ankuriitg:
```suggestion
For other cases, SAC will create table related entities as Spark models.
```

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+One exceptional case is HWC - for HWC source and/or sink, SAC will not create table related entities and always refer to Hive table entities via object id.
```
ankuriitg: Why is this an exceptional case? I think this is already covered by the condition that `SparkSession.builder.enableHiveSupport` is set and `hive.metastore.uris` is not empty

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+Known Limitations (Design decision)
+====
+
+> SAC only supports SQL/DataFrame API (in other words, SAC doesn't support RDD).
+
+SAC relies on query listener to retrieve query and examine the impacts. We may not investigate on supporting RDD for future unless strongly necessary.
```
ankuriitg:
```suggestion
SAC relies on query listener to retrieve query and examine the impacts.
```

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+> All "inputs" and "outputs" in multiple queries are accumulated into single "spark_process" entity when there're multple queries running in single Spark session.
+
+There're pros and cons regarding which is used for unit of "spark_process". Currently SAC uses "application ID" as unit - this approach could concentrate overall impacts on the single Spark process for who ran for when.
+
+But we also received other voices on this approach - lineage/relationship graph in "spark_process" are too complicated and confused and less meaningful. It's valid voice, we've filed #261 to investigate changing the unit of "spark_process" entity to query. It doesn't mean we will change it soon. It will be address only if we see the clear benefits on changing it.
```
ankuriitg: Combine the two points above to: 
spark_process maps to an `applicationId` in Spark. This causes lineage/relationship graph in "spark_process" to be complicated and less meaningful. We've filed #261 to investigate changing the unit of "spark_process" entity to query. It doesn't mean we will change it soon. It will be address only if we see the clear benefits on changing it.

README.md
HeartSaVioR: That's not the case. Even it doesn't meet the condition (so we indicate tables as Spark table), source and sink tables for HWC are obviously Hive table.

README.md
```diff
@@ -111,6 +111,66 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
 When Spark application is started, it will transparently track the execution plan of submitted SQL/DF transformations, parse the plan and create related entities in Atlas.
 
+Spark models vs Hive models
+====
+
+SAC classifies table related entities with two different kind of models: Spark / Hive.
+
+We decided to determine which models SAC should create based on whether Hive Hook can take care of table related entities or not. For Hive entities, SAC relies on HMS Atlas hook as the source of truth and expects HMS atlas hook to create it.
+
+As SAC cannot check whether the condition is properly made, SAC relies on precondition: if Spark session is connected to the "remote HMS" (via thrift), SAC assumes Hive Hook is already set up in remote HMS and Hive Hook will take care of entities creation.
```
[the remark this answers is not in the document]
HeartSaVioR: Now it's redundant and I agree below points provides better understanding. Will remove this.

README.md
HeartSaVioR: If we don't describe pros for `applicationId` as id of `spark_process` whereas we describe cons here, it obviously means it has clear benefits on changing and we should just do it. I think we should describe pros and cons of both for rationalization on current status.

README.md
ankuriitg: But can we use HWC if `SparkSession.builder.enableHiveSupport` is NOT set or `hive.metastore.uris` is empty?

README.md
ankuriitg: I don't understand from your message what is the pro of that? Can you please explain?

README.md
```diff
@@ -111,6 +111,64 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+> All "inputs" and "outputs" in multiple queries are accumulated into single "spark_process" entity when there're multple queries running in single Spark session.
+
+There're pros and cons regarding which is used for unit of "spark_process". Currently SAC uses "application ID" as unit - this approach could concentrate overall impacts on the single Spark process for who ran for when.
+
+But we also received other voices on this approach - lineage/relationship graph in "spark_process" are too complicated and confused and less meaningful. It's valid voice, we've filed #261 to investigate changing the unit of "spark_process" entity to query. It doesn't mean we will change it soon. It will be address only if we see the clear benefits on changing it.
+
+> Only part of inputs are tracked in Streaming query.
+
+This is from design choice on "trade-off": Kafka source supports subscribing with "pattern" and SAC cannot enumerate all matching existing topics, or even all possible topics (even it's possible it doesn't make sense, though).
```
ankuriitg:
```suggestion
This is from design choice on "trade-off": Kafka source supports subscribing with "pattern" and SAC cannot enumerate all matching existing topics, or even all possible topics (even if it was possible, it won't make sense).
```

README.md
```diff
@@ -111,6 +111,64 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+We found "executed plan" provides actual topics which each (micro) batch read and processed, but as a result, only inputs which participate on (micro) batch are being included as "inputs" in "spark_process" entity.
```
ankuriitg:
```suggestion
"executed plan" provides actual topics which each (micro) batch reads and processes, and as a result, only inputs which participate in (micro) batch are included as "inputs" in "spark_process" entity. 
```

README.md
```diff
@@ -111,6 +111,64 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+But we also received other voices on this approach - lineage/relationship graph in "spark_process" are too complicated and confused and less meaningful. It's valid voice, we've filed #261 to investigate changing the unit of "spark_process" entity to query. It doesn't mean we will change it soon. It will be address only if we see the clear benefits on changing it.
+
+> Only part of inputs are tracked in Streaming query.
+
+This is from design choice on "trade-off": Kafka source supports subscribing with "pattern" and SAC cannot enumerate all matching existing topics, or even all possible topics (even it's possible it doesn't make sense, though).
+
+We found "executed plan" provides actual topics which each (micro) batch read and processed, but as a result, only inputs which participate on (micro) batch are being included as "inputs" in "spark_process" entity. 
+
+If your query are running long enough that it ingests data from all topics, it would have all topics in "spark_process" entity.
```
ankuriitg:
```suggestion
If your query runs long enough that it ingests data from all topics, it will have all topics in "spark_process" entity.
```

README.md
ankuriitg: I don't understand your point about multiple spark applications since you have mentioned that it may be a problem for table entity as well. Then why are we only dropping support for column entities?

README.md
ankuriitg: Can you check with Aadarsh offline?

README.md
```diff
@@ -111,6 +111,64 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+We haven't found difficulty on table level, but once we collect difficulties on making table entity being consistent as well, we might also consider drop supporting DDL and blindly create the entities whenever they're referenced.
+
+This doesn't apply to Hive models, which central remote HMS takes care of DDLs and Hive Atlas Hook will take care of updates.
+
+> SAC doesn't track dropping tables (Spark models).
+
+"drop table" event from Spark only provides db and table name, which is NOT sufficient to create qualifierName - especially we separate two types of tables - spark and hive. 
+
+SAC depends on reading the Spark Catalog to get table information but Spark might already dropped the table when SAC notices the table is dropped, so that is NOT consistent. Due to this technical limitation, we have to drop this feature.
```
ankuriitg:
```suggestion
SAC depends on reading the Spark Catalog to get table information but Spark will have already dropped the table when SAC notices the table is dropped so that will not work. Due to this technical limitation, we have to drop this feature.
```

README.md
HeartSaVioR: Suppose there's a Spark application which specific user ran from 3 days ago, and the application ran three queries.

If we set ID as query ID for `spark_process`, spark_process will show the impact for specific query but we have to manually collect the impacts from three queries to see the impact for the application. Other trade-off case is query on structured streaming: if the query reads from its checkpoint, query ID doesn't change, which would hide the information when query is re-started.

Other than that, maybe we can track YARN application ID (if yarn scheduler is used) much easier than query id, as the value will be used an ID for both YARN application and Spark application, and the value will be listed in YARN application list, JHS, SHS. For query ID, we may need to dig down log file to figure out which application ran the query, to fix up if something got wrong. Otherwise someone should maintain the information for (application ID, query ID) pairs, which should consider the case for structured streaming query as well. Maybe we could consider adding `application ID` to maintain pair in `spark_process` entity if we want to change the id to query ID.

I'm not saying taking `application ID` as ID of `spark_process` is better. I agree having accumulated sources and sinks on relation graph is in many case less meaningful, but changing the ID would be a breaking change, so I just want to be careful.

README.md
HeartSaVioR: According to the README on HWC, it requires HS2 information instead of HMS, so I'd think twos are irrelevant.

https://github.com/hortonworks-spark/spark-llap/blob/master/README.md

README.md
HeartSaVioR: The thing is we don't support it, not because of lack of Atlas, but because of lack of Spark. Even Atlas supports it in any way we don't support it. Actually I feel the question as out of topic.

README.md
HeartSaVioR: OK, to be honest, I'd rather want to drop DDL supports on SAC for same reason, and only blindly create table entity whenever referenced. I'm just careful to drop supporting existing features (say, destructive change) without someone's approval. If we are OK with this, I'll just drop DDL support and change the sentences.

README.md
ankuriitg: What additional information is needed to support "drop table" event? If Atlas supports it, then we should find a way to start supporting it. Can't we rely on Hive Atlas Hook for hive tables and add a different mechanism for spark tables?

README.md
ankuriitg: I don't think HWC can be used if any of those configurations are not set: `SparkSession.builder.enableHiveSupport` or `hive.metastore.uris`. So, I think we can remove this statement.

README.md
HeartSaVioR: I like your suggestion, though I don't think we have a plan to do it. We might have the plan eventually, so that's OK.

README.md
HeartSaVioR: At least document doesn't look like requiring that. I'll try to confirm whether the condition is true.

README.md
HeartSaVioR: Nice summary. Looks great to me. I'll reflect the change.

README.md
HeartSaVioR: First of all, all discussions regarding DDL are for Spark table. For Hive table, SAC never deals with DDL and Hive Atlas Hook will do that. For Spark table, we should let Spark pass not only db and table name but also the snapshot of catalog table just before table is deleted. We would need Spark side patch, so maybe we can add statement that "we're working on it", and may also  start investigating it.

Btw, looks like we have different perspective regarding supporting DDL. I'll comment it again.

HeartSaVioR:
Looks like I need to clarify my perspective on supporting DDL for Spark table.

The point is that SAC can't guarantee consistent view of the entity among multiple spark applications, especially events might be out of order. It would be no matter if SAC only creates entity blindly whenever they've referred - we do for other entities. Only table entities are being tracked separately via external catalog listener.

Based on the fact, would we want to provide as many features as possible with "best-effort" and let end users tolerate inconsistency, or drop supporting these features altogether and just provide what we can do consistently? Latter means dropping DDL support.

README.md
```diff
@@ -111,6 +111,64 @@ When running on cluster node, you will also need to distribute this keytab, belo
...
+
+This is from design choice on "trade-off": Kafka source supports subscribing with "pattern" and SAC cannot enumerate all matching existing topics, or even all possible topics (even if it was possible, it won't make sense).
+
+"executed plan" provides actual topics which each (micro) batch reads and processes, and as a result, only inputs which participate in (micro) batch are included as "inputs" in "spark_process" entity. 
+
+If your query runs long enough that it ingests data from all topics, it will have all topics in "spark_process" entity.
+
+> SAC doesn't support tracking changes on columns (Spark models).
+
+We are investigating how to add support for column entity. The main issue we face is how to make this change consistent when multiple spark applications make changes to the same table/column.
```
ankuriitg: Doesn't Atlas hook for Hive also take care of column entities? Can we verify please?

HeartSaVioR: I expect HMS can safely track the changes via DDL because concurrent modification is not happening there or at least with transactional manner. I expect only one writer for Atlas hook for Hive per any event. If not I'd suspect it.

ankuriitg: I missed that this is only for Spark models. Sounds good then.

HeartSaVioR:
Thanks @ankuriitg and @arunmahadevan for putting lots of efforts on thoughtful reviews! Just merged.

PR #264, opened 2019-06-08
[SAC-263][CORE] Qualify path with Spark default filesystem when scheme is not specified

### LINKED ISSUE #263: Qualify path with Spark default filesystem when scheme is not specified
Spark uses default filesystem when path is given without scheme. For example, if HDFS is not set up in Spark, `/tmp` is interpreted as `file:///tmp` whereas if HDFS is set up in Spark, `/tmp` is interpreted as `hdfs://<hdfs>/tmp`.

SAC should follow up the behavior when recording entity to properly create right type of entity.

HeartSaVioR:

Please refer #263 for more details.

This patch proposes path to be qualified with Spark default filesystem when scheme of the path is not specified.

## How was this patch tested?

Added UT. Also manually tested.

> query

```
import org.apache.spark.sql.types.StructType
val userSchema = new StructType().add("col0", "string").add("col1", "string").add("col2","string")
val df = spark.readStream.schema(userSchema).json("/tmp/stream_messages")
 df.writeStream.format("json").option("checkpointLocation","/tmp/chkpt").option("path","/tmp/hdfs_streams").start()
```

> screenshots

Screen Shot 2019-06-08 at 12 04 49 PM

Screen Shot 2019-06-08 at 12 04 35 PM
Screen Shot 2019-06-08 at 12 04 43 PM

Screen Shot 2019-06-08 at 12 05 01 PM

Without this patch, these entities are created as `fs_path`.

This closes #263

### DIFF
--- a/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
+++ b/spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
@@ -21,15 +21,15 @@ import java.io.File
 import java.net.{URI, URISyntaxException}
 
 import com.hortonworks.spark.atlas.sql.KafkaTopicInformation
-
 import org.apache.atlas.AtlasConstants
 import org.apache.atlas.model.instance.{AtlasEntity, AtlasObjectId}
 import org.apache.commons.lang.RandomStringUtils
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.hive.ql.session.SessionState
 import org.apache.spark.sql.catalyst.catalog.{CatalogDatabase, CatalogTable}
-import com.hortonworks.spark.atlas.{SACAtlasEntityReference, SACAtlasEntityWithDependencies, SACAtlasReferenceable, AtlasUtils}
+import com.hortonworks.spark.atlas.{AtlasUtils, SACAtlasEntityReference, SACAtlasEntityWithDependencies, SACAtlasReferenceable}
 import com.hortonworks.spark.atlas.utils.{JdbcUtils, SparkUtils}
+import org.apache.spark.sql.SparkSession
 
 
 object external {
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
@ankuriitg @arunmahadevan Could you please take a time to review this when you have time? Thanks!

HeartSaVioR:
@ankuriitg @arunmahadevan I'll merge the PR in tomorrow if there's no new feedback today.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -108,23 +108,32 @@ object external {
...
-      if (uri.getScheme() != null) {
-        return uri
-      }
+    val uri = new URI(path)
+    if (uri.getScheme() != null) {
+      return uri
+    }
+
+    val qUri = qualifiedPath(path).toUri
+    if (qUri.getScheme == "file") {
```
arunmahadevan: why this additional check ?

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -108,23 +108,32 @@ object external {
...
-      val uri = new URI(path)
-      if (uri.getScheme() != null) {
-        return uri
-      }
+    val uri = new URI(path)
+    if (uri.getScheme() != null) {
+      return uri
+    }
+
+    val qUri = qualifiedPath(path).toUri
```
arunmahadevan: If its the case that all paths in spark has to be a Hadoop Path then this logic seems OK.
If its possible that in some source/sink a path may not be a Hadoop path (e.g. a java.nio.Path referring to a local file and scheme not specified), but we resolve it as HDFS path when it reaches here (because fs.Defaultfs is hdfs://...) ?

HeartSaVioR: Unless Spark puts extra efforts to take care of both file systems (which I don't think Spark would), Spark also has no idea to determine whether the given path is local or HDFS, and Spark has to depend on `fs.defaultFS`.

Other than that, this patch doesn't break existing UTs, hence the patch guarantees working with the case where path is in file system and scheme is not specified, and fs.defaultFS is not set to `hdfs://`.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
HeartSaVioR: This is from existing code (it looks to be borrowed from Spark) and I wouldn't want to drop the existing code where I'm not 100% sure.

ankuriitg: I also have the same question that why is this additional check added?

HeartSaVioR: Because below code only makes sense for local file, as it retrieves absolute path via leveraging `File`. For HDFS file, we may not even need to do it, as we expect absolute path for HDFS once it qualifies.

That said, if `makeQualified` guarantees to always convert the path as absolute path, we even don't need this. I'll check it.

HeartSaVioR: `makeQualified` looks to convert the path as absolute path, but scheme part looks to be different.

If we apply below logic, `file://` is used as a prefix before path. If we rely on `makeQualified`, `file` is used.

From reference of [wikipedia](https://en.wikipedia.org/wiki/File_URI_scheme), both seem to be valid file URI, but given we have been providing it as `file://`, we might be better to just leave this as it is.

ankuriitg: Can you add this in code comments?

HeartSaVioR: OK. I'll add the comment.

HeartSaVioR: Updated.

spark-atlas-connector/src/main/scala/com/hortonworks/spark/atlas/types/external.scala
```diff
@@ -121,6 +121,11 @@ object external {
     }
 
     val qUri = qualifiedPath(path).toUri
+
+    // Path.makeQualified always converts the path as absolute path, but for file scheme
+    // it provides different prefix.
+    // (It provides prefix as "file" instead of "file://", which both are actually valid.)
+    // Given we have been providing it as "file://", we will keep this as it is.
```
ankuriitg: Maybe change it to: Given we have been providing it as "file://", the logic below changes the scheme of type "file" to "file://"

HeartSaVioR: Thanks for the suggestion. Applied.

HeartSaVioR:
Thanks for reviewing!

